PR #167, opened 2026-07-31
docs: update endpoint resolution plan status to completed (#165)

### LINKED ISSUE #165: endpoint-resolution-precedence-plan.md still says Status: Planned, but the fix already shipped
`documentation/docs/endpoint-resolution-precedence-plan.md` documents an endpoint-precedence bug and marks itself `Status: Planned`. The fix actually shipped in commit `bb0b71a7` ("fix: enforce endpoint precedence and tighten fallback behavior") — `discover_chat_config_from_sbom()` in `nuguard/common/endpoint_probe.py` now explicitly returns the configured endpoint immediately ("Explicit endpoint is authoritative and must never be overridden by SBOM"), and both `behavior` and `redteam` call sites guard on it correctly.

**Fix:** update the doc's `Status` to `Shipped`/`Done` (or fold its content into a changelog entry and archive it) so it stops reading as an open bug report. (Closed #160, which was filed against the now-fixed behavior described in this doc.)

TheRodzz:
Closes #165

### Summary
Updates the status in `documentation/docs/endpoint-resolution-precedence-plan.md` from `Planned` to `Completed` as the endpoint resolution fix has already shipped.

### DIFF
--- a/documentation/docs/endpoint-resolution-precedence-plan.md
+++ b/documentation/docs/endpoint-resolution-precedence-plan.md
@@ -1,6 +1,6 @@
 # Endpoint Resolution Patch Plan (Config Must Win)
 
-Status: Planned
+Status: Completed
 
 Scope: `validate`, `behavior`, and `redteam` endpoint resolution for target chat paths.
 


KanishkThamman (approved):
LGTM!

PR #241, opened 2026-08-10
fix(sbom): normalize Windows paths in deployment-URL discovery (#234)

### LINKED ISSUE #226: [Bug]: package discovery
## Expected Behavior

All packages used by the application used be discovered.

## Current Behavior

If the requirements.txt or package.json file is inside few sub-folders we don't discover it.

## Possible Solution

Scan deeper into the sub-folders for the packages.

## Context

Incomplete package or empty list means false positive sense of security.

### LINKED ISSUE #234: [Bug]: Deployment URL Discovery Fails on Windows Paths in SBOM Generation
## Expected Behavior

When SBOM is generated from a Windows CLI run, deployment URL extraction should not miss GitHub Actions workflow files.

## Current Behavior

When SBOM is generated from a Windows CLI run, deployment URL extraction can miss GitHub Actions workflow files because the deployment-path hint check expects forward slashes while scanned file paths use backslashes. As a result, workflow-based URL harvesting and Azure App Service URL reconstruction may not execute, leaving deployment_urls empty even when deploy metadata exists in workflow content.
The same works in Linux.

## Recommended Solution

•	Normalize scanned file paths to POSIX style before applying deployment-path hint matching.
•	Add separator-agnostic checks for workflow paths (support both / and ).

## Impact

•	deployment_urls may be empty in SBOM for Windows users.
•	Runtime flows that rely on SBOM-derived backend URL fallback can lose automatic target resolution.
•	Users must manually provide backend target URL, increasing setup friction and risk of misconfiguration.

nikhilpatidar:
## PR Type

- [x] Bug fix
- [ ] Feature

Fixes #234 and adds regression coverage for #226.

**Issue #234** — Deployment-URL detection in `application_summary.py` did a case-insensitive substring match on the workflow path, but only against the raw string. On Windows the path uses backslashes, so the deployment-URL hint never matched and the SBOM shipped with no application.summary.deployment_hint. The fix normalizes separators before matching:

```python
normalized_path = path.replace("\\\\", "/")
lower_path = normalized_path.lower()
```

**Issue #226** — Sub-folder manifest discovery (e.g. `apps/api/requirements.txt`) was already working via the recursive `_iter_dependency_files` walker. This PR adds regression tests so it stays that way.

**Tests**
- `tests/sbom/test_application_summary_paths.py` — 5 tests for the Windows-path fix
- `nuguard/sbom/tests/test_subfolder_discovery.py` — 6 tests for the sub-folder walker

Closes #234, Closes #226

### DIFF
--- a/tests/sbom/test_application_summary_paths.py
+++ b/tests/sbom/test_application_summary_paths.py
@@ -1,0 +1,87 @@
+"""Tests for application_summary path handling.
+
+Verifies that deployment-context extraction works regardless of the path
+separator used by the platform that produced the path. On Windows, ``os.walk``
+and ``Path.relative_to`` yield backslash-separated paths, so deployment-file
+hints defined with forward slashes (``.github/workflows/``, ``/k8s/``,
+``infra/``) would silently fail to match and ``deployment_urls`` would end up
+empty even when workflow content clearly contains URLs.
+
+See issue #234.
+"""
+
+from __future__ import annotations
+
+from nuguard.sbom.core.application_summary import extract_deployment_context
+
+
+def _workflow_content(url: str = "https://my-backend.azurewebsites.net") -> str:
+    return f"""name: Deploy
+on: [push]
+jobs:
+  deploy:
+    steps:
+      - run: echo {url}
+"""
+
+
+def _azure_workflow_with_env_var() -> str:
+    return """name: Deploy
+on: [push]
+env:
+  AZURE_WEBAPP_NAME: my-backend
+jobs:
+  deploy:
+    steps:
+      - run: deploy
+"""
+
+
+def test_windows_style_workflow_path_matches_hint() -> None:
+    """A backslash-separated path must still trigger the workflow hint."""
+    files = [
+        (".github\\workflows\\deploy.yml", _workflow_content()),
+    ]
+    result = extract_deployment_context(files)
+    assert "GitHub Actions" in result["deployment_platforms"], (
+        f"Expected GitHub Actions detection on Windows path, got "
+        f"{result['deployment_platforms']!r}"
+    )
+    assert "https://my-backend.azurewebsites.net" in result["deployment_urls"]
+
+
+def test_windows_style_workflow_path_reconstructs_azure_url() -> None:
+    """AZURE_WEBAPP_NAME env-var URL reconstruction must work on Windows paths."""
+    files = [
+        (".github\\workflows\\deploy.yml", _azure_workflow_with_env_var()),
+    ]
+    result = extract_deployment_context(files)
+    assert "https://my-backend.azurewebsites.net" in result["deployment_urls"]
+
+
+def test_windows_style_kubernetes_path_matches_hint() -> None:
+    """A backslash-separated ``k8s`` path must still trigger the Kubernetes hint."""
+    files = [
+        ("k8s\\deployment.yaml", "apiVersion: apps/v1\nkind: Deployment\n"),
+    ]
+    result = extract_deployment_context(files)
+    assert "Kubernetes" in result["deployment_platforms"]
+
+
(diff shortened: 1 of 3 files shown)


tests/sbom/test_application_summary_paths.py
```diff
@@ -0,0 +1,87 @@
...
+def test_windows_style_workflow_path_reconstructs_azure_url() -> None:
+    """AZURE_WEBAPP_NAME env-var URL reconstruction must work on Windows paths."""
+    files = [
+        (".github\\workflows\\deploy.yml", _azure_workflow_with_env_var()),
+    ]
+    result = extract_deployment_context(files)
+    assert "https://my-backend.azurewebsites.net" in result["deployment_urls"]
+
+
+def test_windows_style_kubernetes_path_matches_hint() -> None:
```
KanishkThamman, line 62: Verified by reverting the fix locally: this test still passes without it — it's actually exercised by the content-based `apiVersion:` check, not the path-hint match it claims to cover. Doesn't prove the fix.

tests/sbom/test_application_summary_paths.py
```diff
@@ -0,0 +1,87 @@
...
+    """A backslash-separated ``k8s`` path must still trigger the Kubernetes hint."""
+    files = [
+        ("k8s\\deployment.yaml", "apiVersion: apps/v1\nkind: Deployment\n"),
+    ]
+    result = extract_deployment_context(files)
+    assert "Kubernetes" in result["deployment_platforms"]
+
+
+def test_windows_style_infra_path_matches_hint() -> None:
```
KanishkThamman, line 71: Same issue as the kubernetes-path test above — passes via the `azurerm_app_service` content match regardless of the path normalization fix. Consider a case that only matches via a path hint (e.g. an `/infra/` folder with content that doesn't independently trigger detection).

KanishkThamman (dismissed):
Core fix (backslash normalization) is correct and safe on POSIX CI. Two nits inline on test coverage.

tests/sbom/test_application_summary_paths.py
nikhilpatidar: Added in 8a113c4a. Added test_windows_style_kubernetes_path_extracts_url_via_path_hint which uses path `myapp\\k8s\\frontend.yaml` with content that has no Kubernetes marker (no `apiVersion:`, no other k8s trigger). The URL only flows into deployment_urls because the `/k8s/` path-hint matches after backslash-to-slash normalization. With the fix reverted, the raw backslash path never matches and the test fails.

tests/sbom/test_application_summary_paths.py
nikhilpatidar: Added in 8a113c4a. Added two path-hint tests that use content with no Terraform/Azure markers:
- test_windows_style_infra_path_extracts_url_via_path_hint: `infra\\README.md` with a plain text URL, no `azurerm_` resource.
- test_windows_style_deployment_path_extracts_url_via_path_hint: `app\\infra\\notes.txt` with a plain text URL, also no Azure marker.

Both rely on the `infra/` trailing-slash path hint matching after normalization. With the fix reverted, both fail (the URL is silently dropped). The original vacuous tests on lines 62 and 71 are left in place for backwards compatibility — they still pass and continue to assert the Kubernetes/Azure detection markers on Windows paths, but the new tests are the ones that actually prove the path normalization works.

nikhilpatidar:
@KanishkThamman Addressed both Copilot nits in commit 38aea632 on bug/sbom-extraction-fixes: applied ruff format to nuguard/sbom/tests/test_subfolder_discovery.py so the _touch calls match the rest of the file's multiline style, and dropped the stale PR #241 reference from the section header in tests/sbom/test_application_summary_paths.py.

PR #242, opened 2026-08-10
feat(cli): align --output parent-dir handling and add behavior --sbom (#233)

### LINKED ISSUE #233: [Feature]: Align CLI flag behavior across nuguard commands
## Expected Behavior

1. Update behavior CLI flags to include --sbom.
2. Update analyze and redteam to auto-create parent folder for --output

## Current Behavior

1. Commands such as analyze and redteam accept --sbom flag, however behavior uses only nuguard.yaml for sbom file. 
2. nuguard behavior auto-creates parent folder for --output, while analyze and redteam fail in the same scenario if parent folder does not exist (e.g nuguard analyze --output reports/analysis.md fails, but nuguard behavior --output reports/behavior.md succeeds when reports folder is not present).

nikhilpatidar:
## PR Type

- [ ] Bug fix
- [x] Feature

Fixes #233.

Three CLI commands (`nuguard analyze`, `nuguard redteam`, `nuguard behavior`) needed `--output` to create its parent directory automatically when it didn't exist, and `behavior` was missing an `--sbom` flag consistent with the other commands.

**Changes**
- `nuguard/cli/commands/analyze.py` — create parent dir before writing output
- `nuguard/cli/commands/redteam.py` — create parent dir before writing output
- `nuguard/cli/commands/behavior.py` — add `--sbom` flag, create parent dir before writing output

**Tests**
- `tests/cli/test_output_parent_dir.py` — 5 tests covering each command's parent-dir handling and the behavior --sbom flag

Closes #233

### DIFF
--- a/nuguard/cli/commands/behavior.py
+++ b/nuguard/cli/commands/behavior.py
@@ -27,6 +27,14 @@ def behavior_command(
     config: Optional[Path] = typer.Option(
         None, "--config", "-c", help="Path to nuguard.yaml"
     ),
+    sbom: Optional[str] = typer.Option(
+        None,
+        "--sbom",
+        help=(
+            "Path to AI-SBOM JSON file to seed behavior analysis with. "
+            "Falls back to 'sbom:' in nuguard.yaml when --config is set."
+        ),
+    ),
     mode: str = typer.Option(
         "static+dynamic",
         "--mode",
@@ -196,7 +204,7 @@ async def _run_behavior(
     # 4. Load AI-SBOM and auto-enrich if confidence is low
     sbom = None
     sbom_path_obj: Path | None = None
-    raw_sbom_path = cfg.sbom_path
+    raw_sbom_path = sbom or cfg.sbom_path
     if raw_sbom_path:
         sbom_path_obj = Path(raw_sbom_path)
         try:
--- a/nuguard/cli/commands/behavior.py
+++ b/nuguard/cli/commands/behavior.py
@@ -32,7 +32,8 @@ def behavior_command(
         "--sbom",
         help=(
             "Path to AI-SBOM JSON file to seed behavior analysis with. "
-            "Falls back to 'sbom:' in nuguard.yaml when --config is set."
+            "Falls back to 'sbom:' in nuguard.yaml (located via --config or "
+            "the current working directory)."
         ),
     ),
     mode: str = typer.Option(
--- a/nuguard/cli/commands/behavior.py
+++ b/nuguard/cli/commands/behavior.py
@@ -135,6 +135,7 @@ def behavior_command(
     asyncio.run(
         _run_behavior(
             config_path=config,
+            sbom_override=sbom,
             mode=effective_mode,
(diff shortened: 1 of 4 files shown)


nuguard/cli/commands/behavior.py
```diff
@@ -27,6 +27,14 @@ def behavior_command(
     config: Optional[Path] = typer.Option(
         None, "--config", "-c", help="Path to nuguard.yaml"
     ),
+    sbom: Optional[str] = typer.Option(
```
KanishkThamman, line 30: This flag is never forwarded to `_run_behavior()` — the call site's kwargs don't include it, so the CLI value is discarded.

nuguard/cli/commands/behavior.py
```diff
@@ -196,7 +204,7 @@ async def _run_behavior(
     # 4. Load AI-SBOM and auto-enrich if confidence is low
     sbom = None
     sbom_path_obj: Path | None = None
-    raw_sbom_path = cfg.sbom_path
+    raw_sbom_path = sbom or cfg.sbom_path
```
KanishkThamman: `sbom` here refers to the local `sbom = None` two lines above (the parsed SBOM doc), not the new CLI flag — this always evaluates to `cfg.sbom_path`. Needs a dedicated param (e.g. `sbom_override`) threaded through `_run_behavior()`'s signature and the call site.

KanishkThamman (commented):
The --output parent-dir fix (analyze/redteam) is correct and safe. Holding off approval because the headline feature, `behavior --sbom`, doesn't actually work yet — see inline comments. The two new tests for it pass anyway because their assertions are vacuous (or-chained / exit-code-only), so this shipped silently broken.

nuguard/cli/commands/behavior.py
nikhilpatidar: Fixed in d96b39ae. Threaded into _run_behavior() via a dedicated `sbom_override: Optional[str]` parameter and passed from the call site; the orchestrator now uses `sbom_override or cfg.sbom_path` so the CLI value wins over the config fallback.

nuguard/cli/commands/behavior.py
nikhilpatidar: Fixed in d96b39ae. Renamed the local parsed-SBOM variable's shadowing behavior away from the path: the new `sbom_override` parameter on _run_behavior() is used in `raw_sbom_path = sbom_override or cfg.sbom_path`. The two --sbom tests now assert the literal 'Loaded SBOM: <path>' log line and the resolved path, so they fail (not vacuously pass) if the wiring breaks again.

nikhilpatidar:
@KanishkThamman Follow-up on Copilot's review (no inline findings flagged): commit 32129ee9 on feat/cli-flag-alignment adds the redteam --output parent-dir regression test and clarifies the --sbom help text so the fallback to the config-file sbom key is explicit.

KanishkThamman (approved):
LGTM

PR #244, opened 2026-08-10
fix(policy): clarify 'not found' message for stub show command (#162)

### LINKED ISSUE #162: nuguard policy show always reports 'not found'
`nuguard policy show --policy-id <id>` unconditionally errors with 'not found' — there's no write path anywhere that saves a compiled policy for later lookup by ID (unlike `nuguard sbom register`/`show`, which do have one).

**Fix:** either wire `nuguard policy compile` to also persist to the local DB and implement the lookup, or remove the `show` subcommand until that's built.

nikhilpatidar:
## PR Type

- [x] Bug fix
- [ ] Feature

Fixes #162.

The `nuguard policy show` command is a stub that prints `Policy '<name>' not found` for any input. The misleading part is that the message implies the policy exists but isn't currently loaded — when in reality `show` is not implemented at all. This PR clarifies the message so users know that the command is a stub and points them at `nuguard policy validate` as the working alternative.

**Changes**
- `nuguard/cli/commands/policy.py` — improve the 'not found' stub message; mark `show` as hidden in `--help` so it's not surfaced as a feature

**Tests**
- `tests/cli/test_policy_cli.py` — extended regression tests for the new message and the hidden flag

Closes #162

### DIFF
--- a/nuguard/cli/commands/policy.py
+++ b/nuguard/cli/commands/policy.py
@@ -717,7 +717,7 @@ def _print_assessment_table(assessment: object) -> None:
     _console.print(table)
 
 
-@policy_app.command("show")
+@policy_app.command("show", hidden=True)
 def show(
     policy_id: str = typer.Option(
         ...,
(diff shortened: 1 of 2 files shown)


KanishkThamman (approved):
Small, self-contained, exit code untouched, well tested. LGTM.

Issue #177, opened 2026-08-01
[Go Support] Go AI framework adapters (langchaingo, mcp-go, eino)


KanishkThamman:
## Expected Behavior

NuGuard detects agentic AI framework usage in Go applications and emits
`AGENT`, `TOOL`, `PROMPT`, and `FRAMEWORK` nodes from Go source files, with
the same fidelity as the Python LangGraph adapter
(`nuguard/sbom/adapters/python/langgraph.py`) and MCP server adapter
(`nuguard/sbom/adapters/python/mcp_server.py`).

## Current Behavior

Go applications built with `langchaingo`, `mcp-go`, or ByteDance's `eino`
framework produce no SBOM nodes. The redteam and behavior stages cannot
generate targeted attack scenarios for these applications.

## Possible Solution

Create the following adapters in `nuguard/sbom/adapters/go/`, each as a
`GoFrameworkAdapter` subclass:

1. `langchain.py` — detects `github.com/tmc/langchaingo` imports.
   Emits `AGENT` nodes from `agents.NewOneShotAgent()` /
   `agents.NewConversationalAgent()`, `TOOL` nodes from `tools.Tool` struct
   literals, `MODEL` nodes from `llms.New*()` calls.
   Mirrors `python/langgraph.py` structurally.

2. `mcp_server.py` — detects `github.com/mark3labs/mcp-go`.
   Emits `FRAMEWORK` + `TOOL` nodes from `mcp.NewTool()` calls and
   `server.AddTool()` registrations.
   Mirrors `python/mcp_server.py` structurally.

3. `eino.py` — detects `github.com/cloudwego/eino`.
   Emits `AGENT`/`TOOL`/`MODEL` nodes from component graph construction
   (`eino/compose` patterns).

## Steps to Reproduce (for bugs)

N/A — this is a feature addition.

## Context

Part of #172 (Add Golang Support to NuGuard Pipeline).
Depends on sub-issues 1, 2, and 3.

fatkobra:
Hi @KanishkThamman,

I would be interested in taking this as my next contribution after the
dependencies are ready.

I understand that #177 depends on:

- #173 / PR #196 for Go file discovery and module support;
- #174 / PR #198 for the structured Go parse result;
- #175 for the GoFrameworkAdapter base class.

I will therefore not start an implementation branch until those APIs have
settled, especially the GoParseResult and GoFrameworkAdapter contracts.

Because this issue contains three distinct framework adapters, would you prefer
[...]

Issue #176, opened 2026-08-01
[Go Support] Go LLM client adapters (go-openai, anthropic-sdk-go, google/generative-ai-go)


KanishkThamman:
## Expected Behavior

NuGuard detects direct LLM SDK usage in Go source files and emits correctly
attributed `MODEL` and `FRAMEWORK` nodes — the same node types produced by
`nuguard/sbom/adapters/python/llm_clients.py` and
`nuguard/sbom/adapters/typescript/llm_clients.py` for Python and TypeScript
applications respectively.

## Current Behavior

Go files that use `go-openai`, `anthropic-sdk-go`, or `google/generative-ai-go`
produce no SBOM nodes for the LLM models or provider frameworks being used.

## Possible Solution

Create `nuguard/sbom/adapters/go/llm_clients.py` as a `GoFrameworkAdapter`
subclass:

1. Mirror the provider registry pattern from `typescript/llm_clients.py`
   (`_PROVIDERS` dict mapping provider name → packages + constructor functions).
   Cover: `github.com/sashabaranov/go-openai`, `github.com/anthropics/anthropic-sdk-go`,
   `github.com/google/generative-ai-go`, `github.com/ollama/ollama/api`.
2. Import and use `models_kb.py` (`get_model_details`, `infer_provider`,
   `MODEL_FAMILIES`) — the same shared model knowledge base used by both Python
   and TypeScript adapters. No Go-specific model registry should be created.
3. Detect the `openai.NewClient(config)` proxy pattern (custom `BaseURL`) and
   resolve the provider, mirroring `_resolve_provider_from_base_url()` in
   `python/llm_clients.py`.
4. Emit `MODEL` nodes for `openai.ChatCompletionRequest{Model: "..."}` and
   `FRAMEWORK` nodes for bare client instantiations.

## Steps to Reproduce (for bugs)

N/A — this is a feature addition.

## Context

Part of #172 (Add Golang Support to NuGuard Pipeline).
Depends on sub-issues 1, 2, and 3.

nolanefe:
I’d be happy to take this one. It feels like a good next step after the Go provider/Semgrep work in #223/#231. I see that #176 depends on the Go parsing/base-adapter groundwork in #173–#175, so I’m happy to start once those interfaces are ready, or work against the current direction if you’d prefer me to begin sooner.

KanishkThamman:
[...]

PR #247, opened 2026-08-10
Feat/go framework adapter base

### LINKED ISSUE #175: [Go Support] GoFrameworkAdapter base class
## Expected Behavior

A `GoFrameworkAdapter` base class exists at
`nuguard/sbom/adapters/go/_go_base.py` that all Go framework adapters subclass,
providing Go-appropriate `can_handle()` semantics and shared helper methods —
analogous to `TSFrameworkAdapter` in
`nuguard/sbom/adapters/typescript/_ts_regex.py`.

## Current Behavior

No Go adapter base class exists. If individual Go adapters were written without
a shared base, they would each re-implement `can_handle()` logic and helper
utilities, diverging from the Python/TypeScript pattern and making maintenance
harder.

## Possible Solution

Create `nuguard/sbom/adapters/go/_go_base.py`:

1. Define `GoFrameworkAdapter(FrameworkAdapter)` — subclass of the shared
   `FrameworkAdapter` from `nuguard/sbom/adapters/base.py`. All shared
   fields (`ComponentDetection`, `RelationshipHint`, `_framework_node()`,
   `priority`, `name`) are inherited unchanged.
2. Override `can_handle()` to use substring/prefix matching against full Go
   module paths (e.g. `"github.com/tmc/langchaingo" in mod`), mirroring the
   substring matching that `TSFrameworkAdapter.can_handle()` uses for scoped
   npm packages.
3. Port the shared helpers from `TSFrameworkAdapter` (`_resolve()`, `_clean()`,
   `_fw_node()`, `_template_vars()`) so all Go adapters get the same toolkit
   without duplication.
4. `extract()` contract: receives `GoParseResult` as `parse_result`.

## Steps to Reproduce (for bugs)

N/A — this is a feature addition.

## Context

Part of #172 (Add Golang Support to NuGuard Pipeline).
Depends on sub-issues 1 and 2.
Blocking: all individual Go framework adapter sub-issues.

luqmanbooso:
## What & why

Closes #175

Implements the `GoFrameworkAdapter` base class at `nuguard/sbom/adapters/go/_go_base.py`[cite: 3]. This provides shared extraction utilities and Go-appropriate module path matching semantics for all upcoming Golang framework and LLM client adapters (#176, #177).

---

## Changes Included

* **Package & Base Class Setup**:
  * Created `nuguard/sbom/adapters/go/__init__.py` and `nuguard/sbom/adapters/go/_go_base.py`[cite: 3].
  * Defined `GoFrameworkAdapter` subclassing `FrameworkAdapter` from `nuguard.sbom.adapters.base`[cite: 3].

* **Module Path Matching (`can_handle`)**:
  * Implemented prefix/substring module path matching (e.g., matching `github.com/tmc/langchaingo/llms` against target `github.com/tmc/langchaingo`)[cite: 3].
  * Added support for receiving both `GoParseResult` AST objects and raw import sets/lists[cite: 3].

* **Shared Helper Toolkit**:
  * `_clean(text)`: Strips enclosing quotes (`"`, `` ` ``) and whitespace[cite: 3].
  * `_resolve(val, var_map)`: Resolves variable references to literal values[cite: 3].
  * `_fw_node(file_path, confidence, display_name)`: Emits a standardized `FRAMEWORK` component node with Golang language metadata[cite: 3].
  * `_template_vars(text)`: Extracts `{{ variable }}` prompt template placeholders[cite: 3].

* **Unit Test Suite**:
  * Added unit tests in `nuguard/sbom/tests/test_go_adapter_base.py` verifying `can_handle()` matching, string cleaning, variable resolution, template variable extraction, and `FRAMEWORK` node generation[cite: 4].

---

## How verified

- [x] Executed unit test suite (`pytest -p asyncio nuguard/sbom/tests/test_go_adapter_base.py`).
- [x] Verified linting and formatting (`ruff check nuguard/sbom/adapters/go/`).

---

## Unblocks

* #176 (`[Go Support] Go LLM client adapters`)
* #177 (`[Go Support] Go AI framework adapters`)

### DIFF
--- a/nuguard/sbom/adapters/go/_go_base.py
+++ b/nuguard/sbom/adapters/go/_go_base.py
@@ -1,0 +1,103 @@
+"""Base class for all Golang framework adapters."""
+
+from __future__ import annotations
+
+import re
+from typing import Any
+
+from nuguard.sbom.adapters.base import ComponentDetection, FrameworkAdapter
+from nuguard.sbom.types import ComponentType
+
+_TEMPLATE_VAR_RE = re.compile(r"\{\{\s*(\w+)\s*\}\}")
+
+
+class GoFrameworkAdapter(FrameworkAdapter):
+    """Base class for Go AI framework adapters.
+
+    Subclasses define `module_path` (e.g. "github.com/tmc/langchaingo") or
+    `module_paths` set, and override `extract()`.
+    """
+
+    name: str = "go_base"
+    priority: int = 100
+    module_path: str = ""
+    module_paths: set[str] = set()
+
+    def can_handle(self, imports: set[str] | list[str] | Any) -> bool:
+        """Return True if any import matches the target Go module path(s).
+
+        Supports passing a set/list of import strings, or a GoParseResult object.
+        Substring/prefix matching is used (e.g., "github.com/tmc/langchaingo/llms"
+        matches "github.com/tmc/langchaingo").
+        """
+        if hasattr(imports, "imports"):
+            raw_imports = getattr(imports, "imports", [])
+            import_set = {imp.path if hasattr(imp, "path") else str(imp) for imp in raw_imports}
+        elif isinstance(imports, (set, list)):
+            import_set = set(imports)
+        else:
+            return False
+
+        target_paths = self.module_paths or ({self.module_path} if self.module_path else set())
+        if not target_paths:
+            return False
+
+        for imp in import_set:
+            imp_lower = imp.lower()
+            for target in target_paths:
+                if target.lower() in imp_lower:
+                    return True
+        return False
+
+    # ---------------------------------------------------------------------------
+    # Shared Helper Utilities
+    # ---------------------------------------------------------------------------
+
+    def _clean(self, text: str) -> str:
+        """Strip enclosing quotes and clean whitespace."""
+        cleaned = text.strip()
+        if (cleaned.startswith('"') and cleaned.endswith('"')) or (
+            cleaned.startswith("`") and cleaned.endswith("`")
+        ):
+            return cleaned[1:-1].strip()
+        return cleaned
+
+    def _resolve(self, val: str, var_map: dict[str, str]) -> str:
+        """Resolve a variable reference or return the cleaned literal string."""
+        cleaned = self._clean(val)
+        return var_map.get(cleaned, cleaned)
+
+    def _fw_node(
+        self,
+        file_path: str,
+        confidence: float = 0.95,
+        display_name: str | None = None,
+    ) -> ComponentDetection:
+        """Emit a FRAMEWORK node for this Go adapter."""
+        canonical = f"framework:{self.name}"
+        name = display_name or self.name.replace("_", " ").title()
+        return ComponentDetection(
+            canonical_name=canonical,
+            display_name=name,
+            component_type=ComponentType.FRAMEWORK,
+            adapter_name=self.name,
+            priority=self.priority,
+            confidence=confidence,
+            file_path=file_path,
+            metadata={"language": "golang", "framework": self.name},
+        )
+
+    def _template_vars(self, text: str) -> list[str]:
+        """Extract {{ variable }} placeholder names from Go prompt templates."""
+        return _TEMPLATE_VAR_RE.findall(text)
+
+    def extract(
+        self,
+        code: str,
+        file_path: str,
+        parse_result: Any = None,
+    ) -> list[ComponentDetection]:
+        """Extract SBOM components from Go source code."""
+        if parse_result is None or not self.can_handle(parse_result):
+            return []
+        return [self._fw_node(file_path)]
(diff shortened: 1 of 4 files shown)


fatkobra (changes requested):
Thanks—this is a useful and nicely focused foundation, but I’m requesting changes before merge because this class will define the contract for the remaining Go adapters.

## Blocking

### 1. Use module-path boundary matching in `can_handle()`

The current case-insensitive substring comparison can match unrelated modules such as:

- `github.com/tmc/langchaingo-malicious`
- `github.com/tmc/langchaingo2`
- `evilgithub.com/tmc/langchaingo`

Matching should be case-sensitive and require either an exact module path or a slash-delimited subpackage:

```python
import_path == module_path or import_path.startswith(f"{module_path}/")
```

### 2. Align `_resolve()` and `_clean()` with the actual Go parser output

The Go parser already exposes resolved values through `args` and `positional_args`; it does not provide a `var_map` as part of the parse result. Unresolved identifiers can also appear as values such as `$model`.

Please make `_resolve()` operate on the parser’s instantiation/function-call objects, or otherwise consume their resolved argument mappings directly.

`_clean()` should safely handle the value types the parser can produce, including:

- `None`
- integers and booleans
- lists and dictionaries
- unresolved `$name` references

These values should not cause exceptions or produce false component names.

### 3. Preserve evidence provenance in `_fw_node()`

`_fw_node()` currently loses the matched import’s source line and inherits the default `evidence_kind="regex"`, even though the component was detected from a parsed Go import.

Please pass through the matched import information and emit the appropriate evidence metadata, including:

- the import line number;
- the relevant import snippet or module path;
- an import/AST evidence kind such as `ast_import`.

### 4. Add at least one integration-style test using `parse_go()`

The current dummy parser-shaped objects verify the helper API in isolation, but they do not confirm that the base class works with real `GoParseResult`, `GoImport`, `GoInstantiation`, and parser-produced argument values.

Please add tests covering:

- an exact module-path match;
- a valid `/subpackage` match;
- rejection of look-alike module names;
- case-sensitive matching;
- resolved constants from real parsed Go code;
- unresolved `$name` values;
- non-string argument values;
- preservation of the import line and evidence kind.

## Non-blocking design feedback

- I would prefer reusing the inherited `handles_imports` convention instead of introducing the parallel `module_path` and `module_paths` API. If the new fields are intentional, please document why Go adapters need a separate convention.
- `module_paths` should use an immutable class-level value such as a tuple or `frozenset`, rather than a mutable `set`.
- The behavior when both `module_path` and `module_paths` are set should be explicit; the current fallback pattern can silently ignore one of them.
- Consider keeping `extract()` abstract instead of having the base class silently return a generic framework component. A subclass that forgets to implement extraction could otherwise appear to work while producing incomplete output. An explicit helper such as `_extract_framework_only()` would make that behavior clearer.
- Please clarify which template syntaxes `_template_vars()` is expected to support. Its current name suggests general template extraction, but the implementation appears limited to simple `{{ name }}` placeholders and does not cover common Go template forms such as `{{ .Name }}`.

Once the four blocking items are addressed, the base should be in a much safer position for the downstream adapter work.

nuguard/sbom/adapters/go/_go_base.py
```diff
@@ -0,0 +1,103 @@
...
+            return False
+
+        target_paths = self.module_paths or ({self.module_path} if self.module_path else set())
+        if not target_paths:
+            return False
+
+        for imp in import_set:
+            imp_lower = imp.lower()
+            for target in target_paths:
+                if target.lower() in imp_lower:
```
KanishkThamman, line 48: Nit: naive substring match with no path-segment boundary — target `github.com/sashabaranov/go-openai` matches import `.../go-openai-fork/v2`. Worth anchoring on a `/`-boundary before #176/#177 build real adapters on this.

KanishkThamman (approved):
Not wired into the registry/extractor anywhere, so this is dead code today and safe to merge as scaffolding. One nit inline worth fixing before real adapters build on it.

KanishkThamman:
@luqmanbooso could you take a look at @fatkobra comments since I think he brings up some good points.

luqmanbooso:
Okay, for sure

KanishkThamman:
@luqmanbooso I am assuming that you are working on the comments @fatkobra outlined and thus keeping this PR open for now. Let me know about any new developments you have.

KanishkThamman:
And also @luqmanbooso, do you mind  contacting me at kanishk@nuguard.ai so we can set up a call with you all and the rest of the team to discuss the Go features being implemented? I would be happy to share more information about the meeting and its participants via email. Currently inviting all the contributors for this feature since it's critical

KanishkThamman:
Closing in favor of #299, which supersedes this PR (takeover of the Go framework adapter base). #299 also avoids the substring-matching bug found in this PR's `can_handle`.

PR #249, opened 2026-08-10
feat(sbom): add 'Cloud Run' deployment platform marker (#220)

### LINKED ISSUE #220: [Feature]: Extending AI-BOM support
## Expected Behavior

- Make nuguard discovery more robust for the supported languages and frameworks.
- Ensure AI-BOM is sufficiently rich so the downstream security workflows can be successful without requiring full access to the source code.

## Current Behavior

- some of the application developers have reported that the GCP support is in-sufficient. nuguard does not discover relevant configuration details from manifest files, github actions, or other GCP related configurations inside the application repo.
- it's possible the issue exists for AWS and Azure.

## Possible Solution

- run nuguard across cookbooks and coding samples of various supported languages (Python, JS/TS, Go) and ensure the discovery is accurate and AI-BOM has sufficient details.
- nuguard AI-BOM's first priority is to get the application's AI Stack correct along with all the relationships between the entities (e.g. framework -> agents -> tools -> models -> prompts, etc.). Second priority is to get the cloud services and the application API endpoints correct along with the schema of the APIs.

## Context

- better security validation of an agentic AI application is directly dependent on the rich discovery of the application components (AI stack, Cloud infrastructure, API endpoints, etc.)

nikhilpatidar:
## PR Type

- [x] Bug fix
- [ ] Feature

## What

Adds a more specific `Cloud Run` deployment-platform marker alongside
the existing coarse `GCP` marker. The marker fires when **any** of
these Cloud-Run-specific signals appears in the repository:

- the official `google-github-actions/deploy-cloudrun` GitHub Actions
  step appears in a workflow,
- any `run.googleapis.com` annotation is present,
- a `*.a.run.app` runtime URL appears in the file content, or
- a file path contains the substring `cloudrun`.

The Knative serving API (`apiVersion: serving.knative.dev/v1`) is
**intentionally NOT** a standalone Cloud Run trigger: Knative Serving
is open source and commonly self-hosted on plain GKE/EKS/AKS, so a
manifest using that API alone is not necessarily Cloud Run. It only
contributes to the coarse `GCP` marker (verified by
`test_self_hosted_knative_service_is_not_marked_as_cloud_run`).

## Why

Issue #220 reports that nuguard's GCP discovery is too coarse. The
existing `GCP` marker fires for any GCP-related content (gcloud, Cloud
Functions, Vertex AI, generic `gcr.io` references) without
distinguishing Cloud Run, which is the most common serverless
deployment target for AI apps today. Downstream analyzers and reports
that want to surface Cloud Run-specific concerns (e.g. unauthenticated
invoke URLs, IAM conditions for service-to-service calls) had no
programmatic way to tell those scans apart from generic GCP scans.

## Root cause

`extract_deployment_context` in
`nuguard/sbom/core/application_summary.py` had no Cloud Run-specific
trigger. The coarse `GCP` marker fired on `gcloud`, `cloud run`,
`cloudrun`, `vertex ai`, etc. but those inputs are also common for
non-Cloud-Run GCP services, so a downstream consumer cannot tell which
deployment target is actually in use.

## Changes

- `nuguard/sbom/core/application_summary.py` — adds a second platform
  block in `extract_deployment_context` that appends `Cloud Run` to the
  platform list when any of the five triggers above fire.
- `tests/sbom/test_cloud_run_detection.py` — 6 new regression tests
  pinning the contract end-to-end through
  `extract_deployment_context`:
  - Knative Service YAML → `Cloud Run` + `GCP` + `Kubernetes`
  - Deduplication when multiple triggers fire
  - GitHub Actions `deploy-cloudrun` step → `Cloud Run` + `GCP` +
    `GitHub Actions`
  - `*.a.run.app` URLs captured as `deployment_urls`
  - File path with `cloudrun` substring triggers the marker
  - **Negative**: plain Kubernetes manifest is NOT marked as
    `Cloud Run` (so we don't over-fire)

## Tests

- `tests/sbom/test_cloud_run_detection.py`: 6 new tests, all passing.
- `tests/sbom/` full suite: 188 tests passing.
- `ruff check` clean.
- `mypy` clean on the touched module.

## Backward compatibility

- The `GCP` marker continues to fire in parallel — Cloud Run is a
  strict subset of GCP, so downstream consumers reading `GCP` keep
  working.
- Platforms are already deduplicated by the existing `_uniq(...)` call
  on the platform list, so a single scan produces at most one
  `Cloud Run` and one `GCP` entry even when multiple triggers match.
- No public function signatures change; only the contents of one
  returned list (`deployment_platforms`) gain an additional possible
  entry.

## What this PR does NOT do (issue #220 remains open)

This PR is the narrow, regression-pinning slice of #220 — adding the
marker and the tests. The following remain on the issue tracker for
follow-up PRs:

* Broader cookbook validation across all three clouds (Python / JS-TS /
  Go AI apps on AWS / Azure / GCP) — the request in #220's "Possible
  Solution" list.
* More sophisticated Cloud Run URL reconstruction from
  `K_SERVICE` / `GOOGLE_CLOUD_PROJECT` env vars (the GitHub Actions
  AWS App Service equivalent is implemented for Azure but not yet for
  Cloud Run).
* GCP-specific configuration scanning (e.g. `cloudbuild.yaml`,
  `app.yaml`).
* AWS and Azure parity improvements for issue #220's complaint.

This PR deliberately does **not** use the `Closes #220` syntax,
because GitHub's auto-close behaviour does not understand the
"partial" qualification; the issue should remain open until the
broader cookbook-validation work is done.

Addresses part of #220


nuguard/sbom/core/application_summary.py
```diff
@@ -360,6 +360,18 @@ def extract_deployment_context(files: Sequence[tuple[str, str]]) -> dict[str, li
             or "google_cloud_project" in text_lower
         ):
             platforms.append("GCP")
+        # Cloud Run is more specific than the generic "GCP" marker above.
+        # Trigger on the Cloud Run service config API or the official
+        # GitHub Actions deploy step (issue #220 — these are the two
+        # most common ways an AI app ships to Cloud Run).
+        if (
+            "serving.knative.dev" in text_lower
```
KanishkThamman: Nit: Knative Serving is open source and commonly self-hosted on plain GKE/EKS/AKS, so this trigger alone isn't Cloud-Run-specific and could mislabel a self-hosted Knative-on-GKE deployment as "Cloud Run". Worth a negative test case.

KanishkThamman (dismissed):
Additive, no schema impact. One nit inline on a false-positive trigger.

nuguard/sbom/core/application_summary.py
nikhilpatidar: Fixed in 20c00e3f. Split the Cloud Run trigger into a weak signal (the Knative service API) and a set of strong signals (deploy-cloudrun GitHub Action, `run.googleapis.com` annotation, `*.a.run.app` URL, 'cloudrun' in path). The marker now fires only when a strong signal is present, OR when the Knative service API appears together with a `run.googleapis.com` / `*.a.run.app` co-signal.

Added two tests to test_cloud_run_detection.py:
- test_self_hosted_knative_service_is_not_marked_as_cloud_run: bare Knative Service YAML without any co-signal must NOT be marked Cloud Run (the exact regression).
- test_knative_service_with_run_app_url_is_marked_as_cloud_run: Knative Service + a `*.a.run.app` URL in a separate file IS marked Cloud Run (pins the co-signal path).

All 6 existing tests still pass — the original positive case already includes `run.googleapis.com` alongside `serving.knative.dev`, so it satisfies the new co-signal requirement.

KanishkThamman:
@nikhilpatidar, could you review the code comments and fix them as needed?

nikhilpatidar:
> @nikhilpatidar, could you review the code comments and fix them as needed?

Follow-up on Copilot's review (no inline findings flagged): commit 2d67fd7a on bug/gcp-cloud-run-detection makes run.googleapis.com and .a.run.app standalone Cloud Run triggers (rather than GCP-only) and makes any Cloud Run signal imply GCP. Added two regression tests in tests/sbom/test_cloud_run_detection.py: test_run_app_url_alone_marks_cloud_run_and_implies_gcp and test_run_googleapis_api_annotation_alone_marks_cloud_run.

KanishkThamman:
@nikhilpatidar could you take a look at the suppressed comments

nikhilpatidar:
@KanishkThamman Addressed the suppressed comments on the Cloud Run detection tests in commit 4884b63e on bug/gcp-cloud-run-detection: the module docstring now reflects the actual contract — Knative serving.knative.dev alone is intentionally NOT a Cloud Run trigger (requires a run.googleapis.com / *.a.run.app co-signal), and the `cloudrun` marker matches the substring anywhere in a path (not just file names). The stale PR-description bullet about Knative-alone is a PR-body wording issue only; the code and tests already implement the intended behavior.

KanishkThamman (commented):
Reviewed. Code itself is sound — verified `nuguard/sbom/tests/test_cloud_run_detection.py` (10/10) and the full sbom suite (222 passed, 1 skipped), ruff/mypy clean.

Contradiction with description: the "What" section lists bare Knative YAML (`apiVersion: serving.knative.dev/v1`) as a standalone trigger for the Cloud Run marker. The shipped code explicitly does NOT do this — there's an inline comment ("intentionally NOT a Cloud Run trigger," citing reviewer feedback from issue #220) and a dedicated test (`test_self_hosted_knative_service_is_not_marked_as_cloud_run`) pinning the opposite behavior. Looks like the description is stale from before that feedback was incorporated — worth updating the PR body so the changelog/release notes don't overclaim what's detected.

Nits (non-blocking): a redundant `if "GCP" not in platforms` guard that's a no-op since the list gets deduped later anyway; unanchored `"cloudrun" in text` substring matching could false-positive on e.g. "cloudrunner" (pre-existing pattern, not introduced here).

Please update the PR description to match the actual (narrower) trigger set before merge.

nikhilpatidar:
@KanishkThamman Updated the PR description — the `## What` section now matches the shipped (narrower) trigger set: the Knative service API is listed as `intentionally NOT` a standalone Cloud Run trigger (pointing at `test_self_hosted_knative_service_is_not_marked_as_cloud_run`), and the actual Cloud Run signals (deploy-cloudrun GH Action, `run.googleapis.com`, `*.a.run.app`, `cloudrun` path substring) are the only triggers listed. Changelog/release notes won't overclaim.

PR #252, opened 2026-08-11
fix(sbom): reject argument-injection in --from-repo --ref (git clone)

### LINKED ISSUE #253: Hardening: validate --ref in 'nuguard sbom generate --from-repo' against git argument injection
\`nuguard sbom generate --from-repo <url> --ref <ref>\` passes the user-supplied
\`--ref\` value straight to \`git clone\` as a positional argument. A hostile ref
starting with \`-\` can be reinterpreted by git as a flag (\`--upload-pack=<cmd>\`,
\`--config=<key>=<value>\`, etc.). This is the same argument-injection class
documented in CVE-2017-1000117.

Realistic delivery vector: a copy-pasted command from a malicious README, or
documentation that encourages the operator to run a specific \`--ref\`. Because
NuGuard is itself a security tool that operators run on potentially-untrusted
inputs, hardening this boundary reduces operator-machine attack surface.

The same issue exists in the public \`AiSbomExtractor.extract_from_repo\` API,
where the \`url\` value can carry the same injection class (e.g.
\`--upload-pack=evil\`).

## Expected behavior

* \`nuguard sbom generate --from-repo X --ref --upload-pack=evil\` exits with a
  clear error and never invokes git.
* \`AiSbomExtractor.extract_from_repo(url=\"--upload-pack=evil\", ref=\"main\")\`
  raises \`ValueError\` before starting a subprocess.

## Current behavior

\`git clone --depth 1 --branch <ref> <url> <dest>\` is invoked verbatim. A
\`ref\` or \`url\` beginning with \`-\` is interpreted by git as an option.

## Possible solution

* Reject refs that begin with \`-\`, are empty, or contain whitespace.
* Reject urls that lack an http(s)/ssh scheme or begin with \`-\`.
* Insert a \`--\` separator before the url + dest so any future regression in
  the regex cannot re-introduce the same class (defence in depth).

nikhilpatidar:
## PR Type

- [x] Bug fix
- [ ] Feature

## What

Hardens `AiSbomExtractor._clone_repo` against argument-injection via
`--from-repo --ref` so a hostile ref string (e.g. one pasted from a
malicious README) cannot be reinterpreted by `git clone` as a flag
(`--upload-pack=<cmd>`, `--config=...`, etc.).

Closes #253

## Why

NuGuard is a security tool whose `sbom generate --from-repo` flow
hands a user-supplied `--ref` value straight to `git clone` as a
positional argument. The same git argument-injection class has been
documented since CVE-2017-1000117 and reappears in newer advisories;
modern git rejects most leading-`-` positional arguments but the
protection is not uniform across versions and a hardening check at the
NuGuard boundary costs nothing and removes the operator-machine attack
surface entirely.

The realistic delivery vector is a copy-pasted command or a README
that encourages the operator to run a specific `--ref`, which is a
plausible threat model for a tool that is itself invoked defensively.

## Root Cause

`_clone_repo` assembled its argv with the ref and url as bare
positional arguments:

```python
cmd = ["git", "clone", "--depth", "1", "--branch", ref, url, str(dest)]
subprocess.run(cmd, check=True, capture_output=True)
```

Anything that starts with `-` is parsed by git as an option.

## How

`nuguard/sbom/extractor/core.py`

* Added two private regex constants (`_SAFE_REF_RE`,
  `_SAFE_URL_RE`) at class level.
* `_clone_repo` now validates `ref` and `url` and raises
  `ValueError` with a clear message before any subprocess is
  started.
* Added a `--` separator before `url` and `dest` in the argv so
  that any future regression in the regex cannot re-introduce the same
  injection class (defence in depth).

`tests/sbom/test_clone_repo_arg_injection.py` (new)

21 regression tests covering:

* 8 hostile ref patterns rejected (leading `-`, embedded NUL/CR/LF/TAB,
  whitespace-only, empty) — each verifies `subprocess.run` is **not**
  invoked.
* 7 hostile url patterns rejected (leading `-`, missing scheme,
  `file://`, `git://`, empty/whitespace).
* 4 well-formed urls (`https`, `http`, `ssh`) reach `subprocess.run`
  with the expected argv shape.
* Pins the `--` separator placement in the argv so the
  defence-in-depth check cannot silently regress.

The fix does not change any existing API behaviour for legitimate
inputs — every previously-valid ref and url still produces an identical
argv (plus the new `--` separator, which git treats as an end-of-options
marker that affects nothing when subsequent args are non-flags).

## Tests

```
uv run pytest tests/sbom/test_clone_repo_arg_injection.py -v
# 21 passed in 0.04s

uv run pytest tests/sbom/ tests/cli/test_sbom_cli.py -q
# 221 passed, 1 skipped in 1.54s
```

## Validation

```
uv run ruff check nuguard/sbom/extractor/core.py tests/sbom/test_clone_repo_arg_injection.py
# All checks passed!

uv run ruff format --check tests/sbom/test_clone_repo_arg_injection.py
# 1 file already formatted

uv run mypy nuguard/sbom/extractor/core.py
# Success: no issues found in 1 source file
```

The source-file edit is intentionally NOT auto-formatted: `ruff
format` would rewrite unrelated lines in this 3000-line module,
inflating the diff and obscuring the security fix. The diff against
`main` is +43/-1 in the one function plus a new test file.

### DIFF
--- a/nuguard/sbom/extractor/core.py
+++ b/nuguard/sbom/extractor/core.py
@@ -2461,12 +2461,28 @@ class AiSbomExtractor:
     # Accepting such a ref would let a hostile ref string (e.g. one pasted from
     # a malicious README) trick ``git clone`` into invoking other git options
     # such as ``--upload-pack=<command>`` — a known argument-injection vector.
-    _SAFE_REF_RE = re.compile(r"^[^-,\s\x00][^,\s\x00]*$")
+    _SAFE_REF_RE = re.compile(r"^[^-,\s\x00][^,\s\x00]*\Z")
     # Same defence for the URL: even though the CLI validates it as http(s)
     # upstream, ``extract_from_repo`` is a public API callable from any
     # embedding, so we re-validate here to avoid the same injection class.
-    # Only http(s)/ssh transports are accepted at the clone boundary.
-    _SAFE_URL_RE = re.compile(r"^(https?|ssh)://[^\s\x00]*$")
+    # Accepted at the clone boundary:
+    #   - http(s)://host/path
+    #   - ssh://[user@]host[:port]/path
+    #   - scp-style SSH: [user@]host:path  where the path either starts
+    #     with ``/`` (absolute) or with a non-flag character (so a
+    #     hostile provider cannot smuggle a flag through
+    #     ``git@host:--option``). ``extract_from_repo`` historically
+    #     accepted the scp form, so the regex preserves that compatibility
+    #     while still rejecting anything that smells like an injected flag.
+    # The scp path sub-pattern forbids ``-``/``,``/``\s``/``\x00``/``\Z`` at
+    # the start (no leading flag or whitespace) and continues with
+    # safe characters. The host portion also forbids ``-`` so a hostile
+    # user@ portion cannot itself look like an option.
+    _SAFE_URL_RE = re.compile(
+        r"(?:https?|ssh)://[^\s\x00]*\Z"
+        r"|"
+        r"[A-Za-z0-9._-]+@[A-Za-z0-9._-]+:[^-,:\s\x00][^,\s\x00]*\Z"
+    )
 
     @staticmethod
     def _clone_repo(url: str, ref: str, dest: Path) -> None:
--- a/nuguard/sbom/extractor/core.py
+++ b/nuguard/sbom/extractor/core.py
@@ -2453,11 +2453,53 @@ class AiSbomExtractor:
 
         return doc
 
+    # Git rejects positional refs that begin with ``-`` by refusing them as
+    # "ambiguous argument", but only after it has already consumed earlier
+    # options.  Some versions of git also accept leading ``-`` arguments as
+    # flags in older builds (see CVE-2017-1000117 et al.), so we reject the
+    # value up-front rather than rely on the child process's behaviour.
+    # Accepting such a ref would let a hostile ref string (e.g. one pasted from
+    # a malicious README) trick ``git clone`` into invoking other git options
+    # such as ``--upload-pack=<command>`` — a known argument-injection vector.
+    _SAFE_REF_RE = re.compile(r"^[^-,\s\x00][^,\s\x00]*$")
+    # Same defence for the URL: even though the CLI validates it as http(s)
+    # upstream, ``extract_from_repo`` is a public API callable from any
+    # embedding, so we re-validate here to avoid the same injection class.
+    # Only http(s)/ssh transports are accepted at the clone boundary.
+    _SAFE_URL_RE = re.compile(r"^(https?|ssh)://[^\s\x00]*$")
+
     @staticmethod
     def _clone_repo(url: str, ref: str, dest: Path) -> None:
         if shutil.which("git") is None:
             raise RuntimeError("git executable not found on PATH")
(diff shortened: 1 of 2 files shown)


nuguard/sbom/extractor/core.py
```diff
@@ -2453,11 +2453,53 @@ def _extras_framework(node: Any) -> str:
...
 
+    # Git rejects positional refs that begin with ``-`` by refusing them as
+    # "ambiguous argument", but only after it has already consumed earlier
+    # options.  Some versions of git also accept leading ``-`` arguments as
+    # flags in older builds (see CVE-2017-1000117 et al.), so we reject the
+    # value up-front rather than rely on the child process's behaviour.
+    # Accepting such a ref would let a hostile ref string (e.g. one pasted from
+    # a malicious README) trick ``git clone`` into invoking other git options
+    # such as ``--upload-pack=<command>`` — a known argument-injection vector.
+    _SAFE_REF_RE = re.compile(r"^[^-,\s\x00][^,\s\x00]*$")
```
KanishkThamman: Nit: `$` matches before a trailing newline in Python regex, so `"main\n"` slips past this (verified). Use `\Z` or `re.fullmatch` to close it.

nuguard/sbom/extractor/core.py
```diff
@@ -2453,11 +2453,53 @@ def _extras_framework(node: Any) -> str:
...
+    # Same defence for the URL: even though the CLI validates it as http(s)
+    # upstream, ``extract_from_repo`` is a public API callable from any
+    # embedding, so we re-validate here to avoid the same injection class.
+    # Only http(s)/ssh transports are accepted at the clone boundary.
+    _SAFE_URL_RE = re.compile(r"^(https?|ssh)://[^\s\x00]*$")
```
KanishkThamman: Nit: same `$` vs `\Z` issue as the ref regex above. Also, this rejects scp-style SSH URLs (`git@host:path`), which were previously accepted by `extract_from_repo` — a behavior regression for existing callers using that form.

KanishkThamman (dismissed):
Confirmed this closes a real argument-injection RCE via `git clone --upload-pack=...` — reproduced it locally, the `--` separator + scheme check fixes it. Two small nits inline, not blockers.

nuguard/sbom/extractor/core.py
nikhilpatidar: Fixed in 241cb911. Switched `^[^-,\s\x00][^,\s\x00]*$` to `^[^-,\s\x00][^,\s\x00]*\Z` so a trailing newline in the ref can no longer bypass the regex. The '`main\n` slips past' failure mode is closed. Added a 'main\n' regression case to the parametrize set.

nuguard/sbom/extractor/core.py
nikhilpatidar: Fixed in 241cb911. Extended _SAFE_URL_RE to accept scp-style SSH (`git@host:path`) — the form used by every GitHub/GitLab/Bitbucket SSH clone and historically accepted by extract_from_repo. The path sub-pattern still forbids `-`, `,`, `:, whitespace, and null at the start, so a hostile provider cannot smuggle `git@host:--upload-pack=evil` through.

Added three well-formed scp URLs to test_clone_repo_accepts_well_formed_urls and a new test_clone_repo_rejects_hostile_scp_urls parametrize set for the security boundary.

KanishkThamman (approved):

PR #257, opened 2026-08-11
fix(config): honor endpoint: alias in redteam/behavior override blocks

### LINKED ISSUE #256: [Bug]: redteam/behavior override blocks ignore endpoint: alias (only target_endpoint: honored)
## Expected Behavior

When a user writes `endpoint:` under the `redteam:` or `behavior:` block in `nuguard.yaml` to override a shared `target.endpoint:`, the override should take effect — the same way `endpoint:` and `target_endpoint:` are already accepted as aliases in the shared `target:` block.

## Current Behavior

The `redteam:` override block (in `nuguard/config.py` `_flatten_yaml`) only checks `target_endpoint:`. The `endpoint:` alias is silently ignored, so the shared `target.endpoint:` value wins instead of the override. The same bug exists for `behavior.endpoint:`.

This is asymmetric: the shared block accepts both `endpoint:` and `target_endpoint:` as aliases, but the override blocks accept only `target_endpoint:`. Users who follow `nuguard.yaml.example` line 47 (`# endpoint: /api/chat`) and write `redteam.endpoint: /api/redteam` see their override silently dropped.

## Steps to Reproduce

```yaml
target:
  url: http://shared.test
  endpoint: /api/chat
redteam:
  endpoint: /api/redteam
```

Expected: `target_endpoint` resolves to `/api/redteam`.
Actual: `target_endpoint` resolves to `/api/chat` (shared value wins).

Equivalent reproduction:

```bash
uv run python -c "
import yaml
from nuguard.config import _flatten_yaml
print(_flatten_yaml(yaml.safe_load('''
target:
  url: http://shared.test
  endpoint: /api/chat
redteam:
  endpoint: /api/redteam
'''))['target_endpoint'])
"
# prints '/api/chat' (bug)
```

## Possible Solution

In `_flatten_yaml`, accept `endpoint:` as an alias in the `redteam:` override block (an `elif "endpoint" in redteam:` branch next to the existing `target_endpoint` check), and pre-merge `behavior.endpoint:` into `behavior.target_endpoint` before the shared-as-base merge so the override takes effect.

## Context

`nuguard.yaml.example` line 47 documents `endpoint:` as the canonical name. The existing test `test_redteam_endpoint_overrides_shared_endpoint` only exercises the `target_endpoint:` form, so the alias case was never covered. There is no issue or PR open for this; a `bug/*` branch from `main` implements and tests the fix.

## Environment

- NuGuard version / commit: f5b131bb (current main)
- Python: 3.13
- OS: n/a
- Relevant config: `nuguard/config.py` `_flatten_yaml`, `nuguard.yaml.example` line 47

nikhilpatidar:
## PR Type

- [x] Bug fix
- [ ] Feature

Fixes #256

## What

- Accept `endpoint:` as an alias in the `redteam:` override block in `nuguard/config.py` `_flatten_yaml`, mirroring the shared `target:` block's alias handling.
- Accept `endpoint:` as an alias in the `behavior:` override block by pre-merging `behavior.endpoint:` into `behavior.target_endpoint` before the shared-as-base merge.
- Added 2 regression tests in `tests/test_config.py` (`test_redteam_endpoint_alias_overrides_shared_endpoint`, `test_behavior_endpoint_alias_overrides_shared_endpoint`) pinning the contract for both aliases.

## Why

The shared `target:` block in `nuguard.yaml` accepts both `endpoint:` and `target_endpoint:` as aliases (`nuguard/config.py:199-203`). `nuguard.yaml.example` line 47 documents `endpoint:` as the canonical name. The override blocks (`redteam:` and `behavior:`) only accepted `target_endpoint:`, so users who wrote `redteam.endpoint:` or `behavior.endpoint:` saw their override silently dropped and the shared value win instead. This is an asymmetric alias handling bug — the override blocks must mirror what the shared block accepts.

## Root Cause

- Redteam override block at `nuguard/config.py:233-234` checked only `target_endpoint`; the `endpoint` alias was not consulted.
- Behavior override block at `nuguard/config.py:397-431` injects shared `target_endpoint` into `_shared_for_behavior` and then merges `b = {**_shared_for_behavior, **b}`. A user-written `behavior.endpoint:` key was never renamed to `target_endpoint:` before the merge, so it never overrode the shared value.

## How

- Reproduced the bug with a 5-line `python -c "..."` against the existing `_flatten_yaml` — confirmed `flat['target_endpoint']` resolved to the shared `/api/chat` value when `redteam.endpoint: /api/redteam` was set.
- Inspected the existing `test_redteam_endpoint_overrides_shared_endpoint` test in `tests/test_config.py:335` — it exercises only the long-form `target_endpoint:` path; the alias path was never covered.
- Added the `elif "endpoint" in redteam:` branch next to the existing `target_endpoint` check in the redteam block, plus a `behavior.endpoint` → `behavior.target_endpoint` rename before the shared-as-base merge.
- Added two regression tests in the existing `TestSharedTargetBlock` class so they sit alongside the existing `test_redteam_endpoint_overrides_shared_endpoint` test that pins the long-form contract.

## Test Steps

- [x] Reproduce the original issue: a yaml containing `redteam.endpoint: /api/redteam` against `target.endpoint: /api/chat` resolves to `/api/chat` (wrong).
- [x] Confirm the issue no longer occurs on this branch: same yaml resolves to `/api/redteam` (correct).
- [x] Run `uv run pytest tests/test_config.py -v` — 36 tests pass (34 existing + 2 new).
- [x] Run `uv run pytest tests/test_config.py tests/cli/ tests/sbom/ -q` — 296 passed, 1 skipped.
- [x] Run `uv run ruff check nuguard/config.py tests/test_config.py` — clean.
- [x] Run `uv run mypy nuguard/config.py` — clean.

## Checks

- [x] `make test` passes for the changed files (`uv run pytest tests/test_config.py tests/cli/ tests/sbom/ -q` — 296 passed, 1 skipped)
- [x] `make lint` passes (`ruff check` + `mypy` on `nuguard/config.py`)
- [x] Added regression tests covering both aliases

## Other Notes

- Branched from `origin/main` (`f5b131bb`), `bug/*` → `main` per `CONTRIBUTING.md`.
- No changes to any existing tests; only two new test methods added in the `TestSharedTargetBlock` class.
- The fix is purely additive in terms of behaviour — the long-form `target_endpoint:` path is unchanged; the alias path now works.

### DIFF
--- a/nuguard/config.py
+++ b/nuguard/config.py
@@ -226,12 +226,19 @@ def _flatten_yaml(data: dict[str, Any]) -> dict[str, Any]:
             if _sa_type == "login_flow" and isinstance(_shared_auth.get("login_flow"), dict):
                 flat["redteam_auth_login_flow"] = _shared_auth["login_flow"]
 
-    # Redteam section — overrides shared target block when keys are present
+    # Redteam section — overrides shared target block when keys are present.
+    # Both `endpoint:` and `target_endpoint:` are accepted as aliases, mirroring
+    # the shared ``target:`` block (which accepts both names). The shared block
+    # already does this on lines 199-203; the override must accept the same
+    # aliases so users who wrote ``endpoint:`` (the canonical form documented
+    # in nuguard.yaml.example line 47) see their override take effect.
     redteam = data.get("redteam", {}) or {}
     if "target" in redteam:
         flat["target_url"] = redteam["target"]
     if "target_endpoint" in redteam:
         flat["target_endpoint"] = redteam["target_endpoint"]
+    elif "endpoint" in redteam:
+        flat["target_endpoint"] = redteam["endpoint"]
     if "chat_payload_key" in redteam:
         flat["redteam_chat_payload_key"] = redteam["chat_payload_key"]
     if "chat_payload_list" in redteam:
@@ -419,7 +426,12 @@ def _flatten_yaml(data: dict[str, Any]) -> dict[str, Any]:
                     )
                 if "auth" in shared_target and isinstance(shared_target["auth"], dict):
                     _shared_for_behavior.setdefault("auth", shared_target["auth"])
-                # behavior.* wins — merge shared as the lower-precedence base
+                # behavior.* wins — merge shared as the lower-precedence base.
+                # `endpoint:` and `target_endpoint:` are aliases in the shared
+                # block; mirror that here so a user who writes `behavior.endpoint:`
+                # to override the shared endpoint actually wins the merge.
+                if "endpoint" in b and "target_endpoint" not in b:
+                    b["target_endpoint"] = b["endpoint"]
                 b = {**_shared_for_behavior, **b}
             flat["behavior_config"] = b
 
--- a/nuguard/config.py
+++ b/nuguard/config.py
@@ -232,13 +232,15 @@ def _flatten_yaml(data: dict[str, Any]) -> dict[str, Any]:
     # already does this on lines 199-203; the override must accept the same
     # aliases so users who wrote ``endpoint:`` (the canonical form documented
     # in nuguard.yaml.example line 47) see their override take effect.
+    # Precedence matches the shared block: ``endpoint`` wins over
+    # ``target_endpoint`` when both are set in the same block.
     redteam = data.get("redteam", {}) or {}
     if "target" in redteam:
         flat["target_url"] = redteam["target"]
-    if "target_endpoint" in redteam:
-        flat["target_endpoint"] = redteam["target_endpoint"]
-    elif "endpoint" in redteam:
+    if "endpoint" in redteam:
         flat["target_endpoint"] = redteam["endpoint"]
+    elif "target_endpoint" in redteam:
+        flat["target_endpoint"] = redteam["target_endpoint"]
     if "chat_payload_key" in redteam:
         flat["redteam_chat_payload_key"] = redteam["chat_payload_key"]
     if "chat_payload_list" in redteam:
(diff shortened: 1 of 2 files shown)


nuguard/config.py
```diff
@@ -226,12 +226,19 @@ def _flatten_yaml(data: dict[str, Any]) -> dict[str, Any]:
...
+    # the shared ``target:`` block (which accepts both names). The shared block
+    # already does this on lines 199-203; the override must accept the same
+    # aliases so users who wrote ``endpoint:`` (the canonical form documented
+    # in nuguard.yaml.example line 47) see their override take effect.
     redteam = data.get("redteam", {}) or {}
     if "target" in redteam:
         flat["target_url"] = redteam["target"]
     if "target_endpoint" in redteam:
         flat["target_endpoint"] = redteam["target_endpoint"]
+    elif "endpoint" in redteam:
```
KanishkThamman: Nit: precedence is inverted vs. the shared `target:` block, which prefers `endpoint` over `target_endpoint` (`if endpoint / elif target_endpoint`). Here `target_endpoint` wins over `endpoint`. Only matters if both keys are set in the same block, but worth making consistent.

KanishkThamman (dismissed):
Correct fix for the alias asymmetry, well tested. One nit inline on precedence consistency.

nuguard/config.py
nikhilpatidar: Fixed in d0a94d53. Flipped the precedence in both the redteam and behavior override blocks so `endpoint` (canonical) wins over `target_endpoint` when both keys are set in the same block, matching the shared `target:` block on lines 199-203.

Added 4 new tests to tests/test_config.py::TestSharedTargetBlock:
- test_redteam_endpoint_wins_over_target_endpoint_when_both_set
- test_behavior_endpoint_wins_over_target_endpoint_when_both_set
- test_redteam_target_endpoint_used_when_endpoint_absent (alias resolution still works)
- test_behavior_target_endpoint_used_when_endpoint_absent (alias resolution still works)

Negative-tested: with the precedence reverted, the alias value wins instead of the canonical on both blocks.

KanishkThamman (approved):

PR #261, opened 2026-08-13
fix(sbom): auto-create --output parent directory in sbom generate

### LINKED ISSUE #263: [Bug]: sbom generate rejects --output when parent directory does not exist
## Expected Behavior

`nuguard sbom generate --source <dir> --output <path>` should write the
SBOM to `<path>`, creating the parent directory if it does not exist —
matching `nuguard behavior` (which auto-creates parents) and the
analyze/redteam alignment in PR #242.

## Current Behavior

The command fails before scanning anything:

```
$ nuguard sbom generate --source ./app --output ./reports/nested/app.sbom.json
Error: Output directory './reports/nested' does not exist.
  Hint: Create the directory first or choose an existing path.
```

`_validate_inputs()` in `nuguard/cli/commands/sbom.py` treats a missing
parent directory as a fatal input error and exits with code 1, while
every other output-writing command creates the parent automatically.

## Steps to Reproduce

1. Run `nuguard sbom generate --source <dir> --output <missing-parent>/out.json`
2. Observe the error and non-zero exit.

## Context

Inconsistent CLI behavior across commands; forces manual directory
creation. No security impact.

nikhilpatidar:
Fixes #263

## PR Type

- [x] Bug fix
- [ ] Feature

## What

`nuguard sbom generate --output <path>` now auto-creates the parent
directory of the output path instead of failing with
"Output directory does not exist".

## Why

`nuguard behavior` already auto-creates parent directories on write, and
PR #242 aligns `analyze`/`redteam` to do the same. `sbom generate` was
the outlier: it rejected any `--output` whose parent directory didn't
exist, forcing users to manually create directories before scanning and
aborting before any scan work happened.

## Root Cause

`_validate_inputs()` in `nuguard/cli/commands/sbom.py` treated a missing
parent directory as a fatal input error and exited with code 1, unlike
the other commands which create the parent on write.

## How

- Replaced the "does not exist" rejection with
  `out_parent.mkdir(parents=True, exist_ok=True)`.
- Kept the existing write-permission check (`os.access(..., os.W_OK)`),
  which now runs against the (possibly newly created) directory.
- Added a regression test that invokes the real CLI via `CliRunner` with
  `--output` into a nested, not-yet-existing directory.

## Test Steps

- [x] Reproduced the original issue on `main`: `sbom generate --output <missing>/out.json` exits 1 with "Output directory does not exist"
- [x] Confirmed the fix: same command now exits 0 and writes the SBOM
- [x] New regression test fails on the pre-fix code, passes with the fix
- [x] `uv run pytest tests/cli/test_sbom_cli.py -q` → 19 passed
- [x] `uv run pytest tests/cli/ -q` → 72 passed
- [x] `uv run pytest nuguard/sbom/tests/ -q` → 786 passed, 1 skipped
- [x] `uv run ruff check nuguard/` passes
- [x] `uv run mypy nuguard/` passes

## Checks

- [x] `make test` passes (`uv run pytest tests/ -v`)
- [x] Added/updated tests covering this change (regression test for bugs, new coverage for features)

## Other Notes

The `ruff format --check` non-conformance on `nuguard/cli/commands/sbom.py`
and `tests/cli/test_sbom_cli.py` is pre-existing on `main` and unrelated
to this change; CI gates on `ruff check` + `mypy` only.

### DIFF
--- a/nuguard/cli/commands/sbom.py
+++ b/nuguard/cli/commands/sbom.py
@@ -118,16 +118,17 @@ def _validate_inputs(
     out_parent = output.parent
     try:
         out_parent.mkdir(parents=True, exist_ok=True)
-            except OSError as exc:
-                _err(
-                    f"Could not create output directory '{out_parent}': {exc}",
-                    "Check the path and directory permissions.",
-                )
-            if not os.access(out_parent, os.W_OK):
-                _err(
-                    f"No write permission to output directory '{out_parent}'.",
-                    "Check directory permissions.",
-                )
+    except OSError as exc:
+        _err(
+            f"Could not create output directory '{out_parent}': {exc}",
+            "Check the path and directory permissions.",
+        )
+    
+    if not os.access(out_parent, os.W_OK):
+        _err(
+            f"No write permission to output directory '{out_parent}'.",
+            ...
+        )
 
     # ── Source directory ──────────────────────────────────────────────────────
     if source is not None:
--- a/nuguard/cli/commands/sbom.py
+++ b/nuguard/cli/commands/sbom.py
@@ -127,7 +127,7 @@ def _validate_inputs(
     if not os.access(out_parent, os.W_OK):
         _err(
             f"No write permission to output directory '{out_parent}'.",
-            ...
+            "Check the directory permissions and try again.",
         )
 
     # ── Source directory ──────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


KanishkThamman (approved):
Adds needed functionality and passes all tests. LGTM

PR #270, opened 2026-08-13
fix(sbom): emit valid urn:uuid CycloneDX serialNumber

### LINKED ISSUE #269: sbom: CycloneDX serialNumber is not a valid urn:uuid — BOM fails schema validation
## Summary

`nuguard sbom generate --format cyclonedx` (and `--format cyclonedx-ext`) emits a CycloneDX 1.6 BOM whose `serialNumber` is not a valid URN.

## Root cause

`AiSbomSerializer.to_cyclonedx` builds the serialNumber by interpolating the SBOM's `schema_version` (a semver, e.g. `"1.5.0"`) plus the generation timestamp in place of a UUID:

```python
"serialNumber": f"urn:uuid:{doc.schema_version}-{doc.generated_at.strftime('%Y%m%dT%H%M%SZ')}",
```

The CycloneDX 1.5/1.6 JSON schema requires `serialNumber` to match `^urn:uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`. The current output (`urn:uuid:1.5.0-20260813T210215Z`) does not.

## Impact

The generated BOM is rejected by CycloneDX validation tooling (e.g. `cyclonedx-cli validate`, OWASP Dependency-Track, SBOM interchange validators). Verified against the official CycloneDX 1.6 JSON schema (bundled with the `cyclonedx-bom` dependency): the standard BOM fails schema validation with exactly one error, on `serialNumber`. The same defect exists in `to_cyclonedx_extended`.

## Expected vs actual behaviour

- **Expected:** `serialNumber` is a valid `urn:uuid:<UUID>`; the BOM passes CycloneDX schema validation.
- **Actual:** `serialNumber` is `urn:uuid:1.5.0-<timestamp>`, and the BOM fails schema validation.

## Suggested fix

Generate a spec-compliant deterministic UUID5 URN derived from the document identity (`schema_version` + `generated_at`), so the serialNumber is schema-valid and stable for the same document. Applied to both `to_cyclonedx` and `to_cyclonedx_extended`.

nikhilpatidar:
## Summary
Fixes an invalid CycloneDX `serialNumber` in the SBOM serializer that caused generated BOMs to fail CycloneDX schema validation.

## Bug fix
- [x] Yes (issue #269)

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## Root Cause
`AiSbomSerializer.to_cyclonedx` and `to_cyclonedx_extended` built the BOM `serialNumber` by interpolating `doc.schema_version` (a semver like `"1.5.0"`) plus a generation timestamp in place of a UUID:

```python
"serialNumber": f"urn:uuid:{doc.schema_version}-{doc.generated_at.strftime('%Y%m%dT%H%M%SZ')}",
```

The CycloneDX 1.5/1.6 JSON schema requires `serialNumber` to match `^urn:uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`.

## Expected vs Actual Behaviour
- **Expected:** `serialNumber` is a valid `urn:uuid:<UUID>`; the BOM passes CycloneDX schema validation.
- **Actual:** `serialNumber` is `urn:uuid:1.5.0-<timestamp>`; the standard BOM fails schema validation with exactly one error, on `serialNumber` (verified against the official CycloneDX 1.6 JSON schema bundled with `cyclonedx-bom`). The BOM is rejected by CycloneDX tooling (e.g. `cyclonedx-cli validate`, Dependency-Track).

## Implementation
Add `_cyclonedx_serial_number(doc)` which derives a deterministic UUID5 URN from the document identity (`schema_version` + `generated_at`), keeping the serialNumber schema-valid **and stable for the same document**. Use it in both the standard and extended serializers.

## Tests
- `test_serial_number_format` — strengthened from a `urn:uuid:` prefix check to a full UUID-URN pattern assertion (the prefix-only check let the invalid value through).
- `test_serial_number_deterministic_for_same_document` — serializing the same document twice yields the same serialNumber.
- `test_serial_number_format_extended` — extended format also emits a valid serialNumber.
- All three fail against the pre-fix implementation; all pass with the fix.
- Full `nuguard/sbom/tests/` suite: 788 passed; `ruff check` and `mypy` clean.
- Re-validated against the official CycloneDX 1.6 schema: **0 errors**.

Fixes #269


KanishkThamman (commented):
Reviewed. Fix matches the description exactly (both `to_cyclonedx` and `to_cyclonedx_extended` call sites updated), 3 new tests pass, ruff/mypy clean.

Minor nit: the new `serialNumber` is deterministic per `(schema_version, generated_at)` via `uuid.uuid5`, rather than random per typical CycloneDX convention. Not a spec violation (spec recommends, doesn't mandate, a fresh UUID per instance), and separate `sbom generate` runs still get different values since `generated_at` differs — just flagging for awareness, no change needed.

Safe to merge as-is.

PR #271, opened 2026-08-13
Fix linked issue check, SBOM endpoint consistency, and improve LLM enrichment

KanishkThamman:
This pull request introduces version 0.8.10 of NuGuard, adding a new high-severity structural rule (NGA-021) for detecting IDOR-prone API endpoints without authorization checks, updating documentation and tests accordingly, and ensuring SBOM topology enrichment is always run for compatibility. It also includes minor workflow and CLI improvements and marks a deprecated CLI subcommand as hidden.

**Security rule enhancements:**

* Added new rule NGA-021: Detects API endpoints with user/tenant-scoped path parameters (IDOR surface) that lack authorization or guardrail protection, with corresponding logic, metadata, and comprehensive tests. [[1]](diffhunk://#diff-9a9c8e7923a0c26ffa74874dafc9b00e76be3efd3a2c105a1bcb738b0942ade2R25-R27) [[2]](diffhunk://#diff-9a9c8e7923a0c26ffa74874dafc9b00e76be3efd3a2c105a1bcb738b0942ade2R1501-R1541) [[3]](diffhunk://#diff-9a9c8e7923a0c26ffa74874dafc9b00e76be3efd3a2c105a1bcb738b0942ade2R1565) [[4]](diffhunk://#diff-9a9c8e7923a0c26ffa74874dafc9b00e76be3efd3a2c105a1bcb738b0942ade2R1690-R1695) [[5]](diffhunk://#diff-9a9c8e7923a0c26ffa74874dafc9b00e76be3efd3a2c105a1bcb738b0942ade2R1878-R1883) [[6]](diffhunk://#diff-0e9e25d535855cc269adf4d87d16197af70eb5c514b93a6f19f8b87cd8f55d5dR39) [[7]](diffhunk://#diff-0e9e25d535855cc269adf4d87d16197af70eb5c514b93a6f19f8b87cd8f55d5dR949-R1029) [[8]](diffhunk://#diff-0e9e25d535855cc269adf4d87d16197af70eb5c514b93a6f19f8b87cd8f55d5dL572-R609)
* Updated NGA-006 to also trigger when `auth_required` is explicitly set to `False`, with new tests for this case. [[1]](diffhunk://#diff-9a9c8e7923a0c26ffa74874dafc9b00e76be3efd3a2c105a1bcb738b0942ade2R642) [[2]](diffhunk://#diff-0e9e25d535855cc269adf4d87d16197af70eb5c514b93a6f19f8b87cd8f55d5dR511-R540)

[...]

PR #273, opened 2026-08-13
fix: cherry-pick PR #271 (main -> develop) with conflict resolved

KanishkThamman:
Cherry-picks PR #271's merge commit (main into develop) as a single commit on
top of develop, with the redteam.py conflict pre-resolved.

The only conflict when merging main into develop was in
`nuguard/cli/commands/redteam.py`: main added a `text` output-format branch
(and its `_render_findings_text` helper), while develop added
`out_path.parent.mkdir(parents=True, exist_ok=True)` before writing. Both are
additive/non-overlapping; this keeps both — mkdir runs once before the format
dispatch, and the text branch is preserved.

Verified the resulting tree is byte-identical to a proper `git merge` of
main into develop with the same resolution applied (via
`git rev-parse HEAD^{tree}`), so this is equivalent to completing #271, just
as one commit instead of a merge commit.

## Test plan
- [x] `python3 -m pytest tests/cli -k redteam -q` — 11 passed
- [x] Resulting tree confirmed identical to a real merge commit with the same fix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "id": "nuguard",
   "name": "nuguard",
-  "version": "0.8.8",
+  "version": "0.8.10",
   "description": "AI application security for Claude — generate an AI Bill of Materials (AI-SBOM), run static analysis, behavioral validation, and adversarial red-team testing for AI agents and LLM-powered applications.",
   "author": {
     "name": "NuGuard",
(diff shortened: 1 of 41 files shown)


KanishkThamman:
Closing in favor of #274, which is a much smaller, self-contained fix touching only the conflicting file (nuguard/cli/commands/redteam.py) instead of pulling in all of main's unrelated drift.

PR #278, opened 2026-08-14
test(common): cover auto-enricher disk cache hit and target_url invalidation

### LINKED ISSUE #279: test(common): no regression coverage for auto-enricher disk cache hit and invalidation
## Summary

The auto-enricher's on-disk cache in `nuguard/common/auto_sbom_enricher.py` has no regression coverage for its two most important behaviors:

1. **Cache hit** — a second call with identical SBOM content + `target_url` + `probe_auth_header` returns the cached `*.enriched.json` artifact (`reasons` includes `"enrichment_cache_hit"`) instead of re-running the enrichment pipeline.
2. **Cache invalidation** — changing `target_url` (or `probe_auth_header`, or the SBOM content) changes the `_enrichment_cache_key` (sha256) and produces a cache miss, so a stale enriched artifact is never served for a different target.

## Why it matters

The cache key is a sha256 of (SBOM content, target_url, probe_auth_header). If the key computation regressed (e.g. `target_url` dropped from the hash), the enricher would silently serve a stale enriched artifact — with the previous target's endpoint probe results — across different targets. No test currently catches this.

## Work

Add regression tests covering the cache-hit branch and the target_url invalidation behavior. The invalidation test must fail against a regression where `target_url` is dropped from the cache-key computation.

nikhilpatidar:
## Summary
Adds regression coverage for the auto-enricher's on-disk cache (`nuguard/common/auto_sbom_enricher.py`): the cache-hit branch and the cache-key invalidation behavior were entirely untested.

## Tests
- `test_common_auto_enrich_serves_cached_artifact_on_second_call` — a second call with the identical SBOM object returns the cached artifact (`reasons` includes `enrichment_cache_hit`) instead of re-running the pipeline.
- `test_common_auto_enrich_cache_invalidated_by_target_url` — changing `target_url` invalidates the cache (cache MISS).

## Regression proof
The invalidation test was verified to **FAIL** against a regression where `target_url` is dropped from the cache-key computation (`sha256` of SBOM content + target_url + probe_auth_header): the cache incorrectly hits with the stale artifact and the assertion `expected cache MISS after target_url change` fails. This proves the test provides real protection against a silent cache-key regression that would serve stale enriched SBOMs across different targets.

## Tests run
- `tests/common/test_auto_sbom_enricher.py` — 10 passed
- `tests/common/` — 256 passed
- `ruff check` — clean

Fixes #279


KanishkThamman (commented):
Reviewed. No production code touched. Verified the regression-proof claim is real by injecting the described bug (dropping `target_url` from the cache key) and confirming the new test catches it exactly as claimed.

Minor: `test_common_auto_enrich_cache_invalidated_by_target_url` isn't fully hermetic — it triggers a real `httpx.AsyncClient` call to `http://other-target` (fails fast here with no egress, but could be slow/flaky in an environment with real DNS resolution). Repo already has `respx` as a dev dependency — worth mocking the probe in a follow-up rather than hitting a live-ish host.

Nit: missing a blank line (PEP8, not enforced since ruff excludes `tests/`) between the new test and the previous one.

Safe to merge as-is.

KanishkThamman:
No code changes and only changes in the test therefore the e2e test fail can be ignored

PR #222, opened 2026-08-07
feat(sbom): add C# framework adapters

fatkobra:
Closes #212

## PR Type

- [ ] Bug fix
- [x] Feature

## What

- Added direct C# LLM SDK detection for:
  - Azure OpenAI
  - OpenAI .NET
  - Anthropic C#
- Added Semantic Kernel detection for:
  - kernel builders and instances
  - registered AI model services
  - plugins
  - `[KernelFunction]` methods
  - planners
  - prompt strings
- Added ML.NET detection for:
  - `MLContext` instances
  - transformation pipelines
  - trainers
  - models produced through `Fit`
- Added ASP.NET Core detection for:
  - attribute-routed controllers
  - Minimal API endpoints
  - HTTP methods and expanded route templates
  - request and response schemas
  - conversational payload fields
  - endpoint authorization
- Added shared C# call-expression helpers.
- Added 19 focused adapter tests.

## Why

NuGuard can parse C# source after #211, but it does not yet recognize the AI frameworks and web surfaces used by .NET applications.

This PR implements the framework adapters requested in #212 and supports the broader C# pipeline work tracked in #170.

## How

- Added a source helper that:
  - masks comments, character literals, and string contents before matching calls
  - preserves source offsets and line numbers
  - handles constructor calls, chained calls, generic arguments, named arguments, and nested argument expressions
  - resolves literal strings, string constants, enum members, `nameof(...)`, and simple `Uri` construction
- Added `CSharpLLMClientsAdapter`.
  - Tracks provider identities through client variables.
  - Emits provider `FRAMEWORK` nodes and concrete `MODEL` nodes.
  - Preserves client class, provider, framework, model metadata, and source evidence.
- Added `CSharpSemanticKernelAdapter`.
  - Detects builder creation without treating unrelated `Build()` calls as Semantic Kernel.
  - Detects registered model services, plugins, planners, kernel functions, and prompts.
- Added `CSharpMLNetAdapter`.
  - Detects ML.NET contexts, transformation pipelines, trainers, and trained models.
  - Avoids treating unrelated `Append` calls as ML.NET pipelines.
- Added `CSharpAspNetCoreAdapter`.
  - Detects AI-relevant controller actions and Minimal API routes.
  - Expands `[controller]` and `[action]` route tokens.
  - Handles `Authorize` and `AllowAnonymous`.
  - Extracts record/property request and response schemas where available.
- Exported the adapters from `nuguard.sbom.adapters.csharp`.
- No runtime dependency was added.

[...]

Issue #213, opened 2026-08-05
[C# Support] Wire C# parsing into the SBOM extraction pipeline


KanishkThamman:
## Expected Behavior

`nuguard sbom generate` recognizes `.cs` source files, runs the C# parser and framework adapters against them, and merges the resulting nodes/edges into the `AiSbomDocument` alongside Python/JS/Go output.

## Current Behavior

`.cs` is not in `AiSbomConfig.include_extensions` (`nuguard/sbom/config.py`), so `extractor/core.py`'s source-file walk (`core.py:2450`) skips C# files entirely — this is the config change that does *not* belong in #210 (dependency scanning walks manifests unconditionally by filename, independent of this setting).

## Possible Solution

1. Add `.cs` (and `.csx` if in scope) to `AiSbomConfig.include_extensions`.
2. Wire the `csharp_parser.py` output and adapters from #211/#212 into `extractor/core.py`'s per-file processing, following the existing pattern for how Python/JS parse results feed into node/edge construction.
3. Confirm merged output round-trips through the existing SBOM schema/merger without special-casing (`nuguard/sbom/schemas/aibom.schema.json`, `test_committed_schema_matches_models`).

## Context

Part of #170. Depends on #210, #211, and #212 all landing first — this is the integration point that makes the C# work actually reachable from the CLI.

fatkobra:
Hi @KanishkThamman — I’ll work on the C# extraction-pipeline integration requested in #213.

The scope will include:

- adding `.cs` and `.csx` to the default AI-SBOM source extensions
- registering the C# framework adapters from #212
- parsing C# source files during the extractor’s per-file processing
- dispatching C# parse results to the C# adapters
- merging C# framework, model, tool, prompt, and endpoint detections into `AiSbomDocument`
- associating C# namespace imports with NuGet dependencies
- verifying nodes, edges, evidence, schema round-tripping, and CycloneDX merger output
- focused integration tests for `.cs`, `.csx`, malformed source, custom extension configuration, and language isolation

Because #213 depends on #212 and #222 is still open, I will prepare this but not submit and just wait.

C# Semgrep rules and full fixture applications remain out of scope and are tracked in #214 and #215.

Issue #170, opened 2026-07-31
Add C# Support to NuGuard Pipeline


KanishkThamman:
## Expected Behavior

NuGuard should be able to process C#/.NET-based AI applications end-to-end through its pipeline:

- `sbom generate` should parse C# project files (`.csproj`, `packages.config`, `Directory.Packages.props`) to detect AI/LLM dependencies (e.g. Semantic Kernel, Azure OpenAI SDK, Microsoft.ML) and produce a valid AI-SBOM.
- `analyze` should run C#-aware static detectors (e.g. detecting unvalidated LLM responses, insecure string interpolation used to construct prompts, missing input sanitization before passing user data to AI APIs).
- `behavior` and `redteam` should work against running .NET AI services (ASP.NET Core, Azure Functions, etc.).
- `report` should surface findings with correct file paths and line numbers from `.cs` source files.

## Current Behavior

NuGuard's SBOM generator and static analysis detectors are currently built around Python-centric dependency formats. C# project files (`.csproj`, `packages.config`) are not recognized, and `sbom generate` produces no meaningful output when pointed at a C# repository.

## Context

C# and the .NET ecosystem are widely used in enterprise AI applications, particularly with Microsoft's Semantic Kernel and Azure OpenAI integrations. Adding C# support allows NuGuard to cover .NET-based AI workloads and serve teams building on the Microsoft AI stack.

fatkobra:
Hi @KanishkThamman,

I would be interested in contributing to C# support while the remaining Go
adapter dependencies are settling.

This issue appears to be an umbrella feature covering several independent
areas: project/dependency discovery, C# source parsing, framework detection,
static security rules, runtime behavior/red-team support, and reporting.

Would you be open to splitting it into smaller child issues, similar to the Go
support work under #172?

I would like to begin with a focused foundational PR for C# project-manifest
and NuGet dependency scanning. My proposed scope would be:

- add the required C# source/project visibility to the SBOM configuration;
- discover `.csproj`, `packages.config`, and
  `Directory.Packages.props` files;
- extract NuGet package names and versions from `PackageReference`,
  `PackageVersion`, and `packages.config` entries;
- emit `PackageDep` records with `pkg:nuget/` PURLs;
- add focused tests for:
  - inline `PackageReference` versions;
  - nested `<Version>` elements;
  - central package management through
    `Directory.Packages.props`;
  - `packages.config`;
  - XML namespaces;
  - malformed or incomplete XML;
[...]

PR #290, opened 2026-08-14
feat(sbom): harden C# framework adapter edge cases

### LINKED ISSUE #260: [C# Support] Harden C# framework adapter edge cases before pipeline wiring
## Expected Behavior

The C# framework adapters should preserve C# invocation, routing, request-shape, provider, model, and plugin semantics across common .NET usage patterns:

* Fluent invocation chains such as `Trainer().Fit(...)` and `CreateBuilder().Build()` should preserve the preceding invocation as the receiver and retain the final assignment target.
* ASP.NET Core controller actions should support:

  * a separate method-level `[Route(...)]` combined with `[HttpPost]`, `[HttpGet]`, and related HTTP attributes
  * absolute action templates beginning with `/` or `~/`, which override the controller route
  * collection-valued conversational fields such as `List<Message>` and `string[]`
  * explicitly body-bound primitive prompts such as `[FromBody] string prompt`
* An unqualified legacy `OpenAIClient` should be classified as Azure OpenAI when `Azure.AI.OpenAI` is imported and the OpenAI namespace is not, even when unrelated providers are also imported.
* Semantic Kernel should:

  * avoid creating a fabricated model node when a model or deployment expression cannot be resolved
  * resolve positional plugin names according to the registration method and overload being used

## Current Behavior

A follow-up review of #222 identified these edge cases:

* `find_calls()` only records the identifier chain immediately before the current `(`. In `Trainer().Fit(...)`, `Fit` therefore loses its receiver and the final assignment target.
* A method-level `[Route("complete")]` is ignored when the HTTP attribute has no inline template.
* Absolute ASP.NET action routes are concatenated with the controller route instead of replacing it.
* `chat_payload_list` is always emitted as `false`, including for array and collection fields.
* Primitive parameters are discarded even when marked `[FromBody]` or named like conversational input.
* Legacy `OpenAIClient` is treated as Azure only when Azure is the sole imported provider.
* An unresolved Semantic Kernel model registration derives a model name from the registration method, producing values such as `AzureOpenAI` that are not model identifiers.
* Semantic Kernel plugin extraction always assumes positional argument zero is the plugin name, although object and prompt-directory registrations place it later.

## Possible Solution

* Parse all call expressions before applying the optional name filter.
* Link a call to the immediately preceding invocation when the expressions are joined by `.` or `?.`.
* Inherit the fluent chain’s assignment target and preserve the entire receiver expression.
* Ensure ML.NET `Fit` processing runs after trainer and transform recognition without treating `Fit` itself as the trainer or transform call.
* Fall back to a method-level `[Route]` when the HTTP attribute has no inline route.
* Treat action routes beginning with `/` or `~/` as absolute.
* Derive `chat_payload_list` from the request-field type.
* Preserve body-bound and conversationally named primitive parameters and synthesize their request schema.
* Classify legacy `OpenAIClient` as Azure when Azure is imported and OpenAI is not.
* Skip unresolved Semantic Kernel model registrations instead of fabricating a model identifier.
* Use registration-specific positional candidates for Semantic Kernel plugin names.
* Add focused regression tests for every behavior above.

## Context

This is a follow-up to #212 and PR #222.

The adapters are currently additive-only and are not yet registered in the normal extraction pipeline. This hardening work should land before #213 wires the adapters into regular AI-SBOM generation.

Out of scope:

* `.cs` and `.csx` source configuration
* adapter registry changes
* extraction-pipeline wiring
* C# Semgrep rules
* end-to-end C# fixture applications
* runtime behavior or red-team changes

fatkobra:
Closes #260 

Follow-up to #222.  
Supports the C# pipeline integration tracked in #213.  
Part of #170.

## PR Type

- [ ] Bug fix
- [x] Feature

## What

This PR hardens the C# framework adapters against edge cases identified during the final review of #222.

It adds support for:

- preserving receivers and assignment targets across fluent invocation chains such as:
  - `Trainer().Fit(data)`
  - `Kernel.CreateBuilder().Build()`
- detecting ML.NET trained models created from inline trainer chains
- ASP.NET Core controller actions with separate method-level `[Route(...)]` attributes
- absolute ASP.NET Core action routes beginning with `/` or `~/`
- collection-valued conversational payloads such as:
  - `List<Message>`
  - `IReadOnlyList<Message>`
  - `string[]`
- body-bound primitive prompts such as `[FromBody] string prompt`
- legacy Azure `OpenAIClient` classification when Azure and unrelated provider namespaces are both imported
- unresolved Semantic Kernel model registrations
- registration-specific Semantic Kernel plugin-name argument positions

Focused regression tests were added for each reviewed case.

## Why

The C# adapters introduced in #222 are currently additive-only, but #213 will register them in the normal AI-SBOM extraction pipeline.

The final review of #222 identified several technically valid edge cases that could otherwise produce:

- missing ML.NET trained-model nodes
- incorrect ASP.NET Core routes
- incorrect chat request shapes
- missed primitive prompt inputs
- incorrect Azure/OpenAI provider metadata
- fabricated Semantic Kernel model identifiers
- incorrect Semantic Kernel plugin names

Addressing these cases before #213 lands keeps the pipeline-wiring PR focused on integration rather than adapter correctness.

### Fluent call-chain handling

Updated the shared C# call-expression helper to:

- parse all call expressions before applying adapter-specific name filters
- link a call to the immediately preceding invocation when separated by `.` or `?.`
- preserve the full chained receiver expression
- inherit the assignment target from the beginning of the fluent chain
- retain accurate snippets and source locations

For example:

```csharp
var model = ml.BinaryClassification.Trainers
    .SdcaLogisticRegression()
    .Fit(data);

### DIFF
--- a/nuguard/sbom/adapters/csharp/_source.py
+++ b/nuguard/sbom/adapters/csharp/_source.py
@@ -86,7 +86,8 @@ def find_calls(
 ) -> list[CSharpCall]:
     """Return call expressions whose simple name is in *names*, if supplied."""
     masked = mask_non_code(source)
-    calls: list[CSharpCall] = []
+    all_calls: list[CSharpCall] = []
+    selected_calls: list[CSharpCall] = []
 
     for match in _CALL_RE.finditer(masked):
         callee = re.sub(
@@ -105,9 +106,6 @@ def find_calls(
         if name in _CONTROL_NAMES:
             continue
 
-        if names is not None and name not in names:
-            continue
-
         open_paren = match.end() - 1
         close_paren = _matching(
             masked,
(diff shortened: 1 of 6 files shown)


KanishkThamman (commented):
Reviewed. All 8 described sub-features are implemented with corresponding regression tests, no scope creep found. Verified: full `nuguard/sbom/tests/` suite (879 passed, 1 skipped), ruff/mypy clean on `nuguard/sbom/adapters/csharp/`.

Findings (non-blocking, worth follow-up before #213 wires this into the pipeline):
- `_source.py`'s `find_calls()` dropped its early-exit filter (`if names is not None and name not in names: continue`) to support the new global chain-parent linking, so every call in a file is now fully parsed regardless of the narrow `names` set callers pass. `_chained_parent` itself is a reverse linear scan over all calls seen so far per call — worst-case O(n²) over call count per file. Untested cost; likely fine for typical files but this scans arbitrary user repos, so worth a note for large/generated C# files.
- `aspnet_core.py`'s new `_request_parameter` logic treats any primitive param named `message`/`prompt`/`query`/etc. as the request-body field whenever `from_body` is true **or** the name matches, regardless of actual binding attribute — so `[FromQuery] string prompt` or an unattributed Minimal-API param (query-bound by default) would also get labeled as body-bound. Only the true `[FromBody]` case is tested. Low impact (mislabels source, doesn't miss detection) but not covered by the description or tests.

Safe to merge as-is.

fatkobra:
Thanks, @KanishkThamman. Understood.

I’ll leave #290 unchanged since the findings are non-blocking and the PR is safe to merge.

I’ll open a focused follow-up issue covering:

- replacing the reverse linear chain-parent lookup with a more scalable indexed lookup
- preserving ASP.NET parameter binding sources so query-, route-, form-, and body-bound primitive parameters are not all represented as request-body fields
- regression and performance-oriented coverage for both cases

I’ll address that follow-up in a separate PR against `develop` before starting the #213 extraction-pipeline wiring.

KanishkThamman (approved):

PR #294, opened 2026-08-14
fix(config): stop unset ${VAR} leaking literal 'None' into headers/app_env

### LINKED ISSUE #293: config: unset ${VAR} in headers/app_env leaks the literal string 'None' into HTTP requests
## Summary

When a value in `nuguard.yaml` uses `${VAR}` interpolation and the variable is not set, `_expand_env_vars` correctly turns the value into `None` — but `_flatten_yaml` then **stringifies that `None` back into the literal string `"None"`** for the header-like dict blocks (`target.headers`, `redteam.headers`, `redteam.app_env`, `customer_profile`), because those comprehensions call `str(v)` without filtering `None` first.

## Impact

The literal `"None"` header value reaches the live HTTP client:

- `nuguard redteam` / `nuguard target verify` send `X-Tenant-Id: None` (or similar) to the target when the env var is unset — a malformed, user-visible request header.
- `resolve_auth_runtime` can even seed the auth identity from a `"None"`-valued header (e.g. `Authorization: None` becomes the derived bearer token), silently breaking authentication.
- `redteam.app_env` and `customer_profile` carry `"None"` into app-launch environment variables.
- `redteam.credentials` and the LLM blocks already filter `None`; the header/app_env blocks are inconsistent with that established pattern.

## Additional finding

`nuguard.yaml.example` documents `target.headers` as applying to **behavior + redteam** (`"Extra HTTP headers added to every request (behavior + redteam)"`), and `_flatten_yaml` injects shared `target.headers` into `behavior_config` — but `BehaviorConfig` had **no `headers` field**, so the value was silently dropped by pydantic and behavior never attached those headers to any request. The behavior runner also never read headers from config.

## Fix (PR to follow)

1. Filter `None` values in all header/app_env flattening comprehensions (matching `credentials`).
2. Add a `headers` field to `BehaviorConfig` and merge it into the behavior client's outbound headers.
3. Harden `resolve_auth_runtime._normalize_headers` to drop `"None"`/`""` header values as a final safety net.

## Tests

- `TestUnsetEnvVarNoneFiltering` (full pipeline: env expansion + flatten) — all fail pre-fix.
- `test_resolve_auth_runtime_drops_none_string_*` — fail pre-fix.
- `test_behavior_runner_merges_config_headers_into_client` — fails pre-fix (`KeyError: 'X-Tenant-Id'`).
- `test_shared_headers_reach_resolved_behavior_config` — fails pre-fix (`AttributeError: 'BehaviorConfig' object has no attribute 'headers'`).

nikhilpatidar:
## Problem

`_expand_env_vars` turns an unset `${VAR}` reference into `None`, but `_flatten_yaml` then **stringifies that `None` into the literal string `"None"`** for the header-like dict blocks (`target.headers`, `redteam.headers`, `redteam.app_env`, `customer_profile`) because the comprehensions call `str(v)` without filtering `None` first (`redteam.credentials` and the LLM blocks already filter it).

**Result:** with `X-Tenant-Id: ${MY_TENANT}` and the var unset, the live target receives `X-Tenant-Id: None`. Worse, `resolve_auth_runtime` can seed the derived auth identity from a `"None"`-valued header (e.g. `Authorization: None`).

**Secondary finding:** `nuguard.yaml.example` documents `target.headers` as `behavior + redteam`, and `_flatten_yaml` injects shared `target.headers` into `behavior_config` — but `BehaviorConfig` had **no `headers` field**, so pydantic silently dropped it and behavior never attached those headers to requests.

## Fix

1. Filter `None` in all header/app_env flattening comprehensions (shared `target.headers`, `redteam.headers`, `redteam.app_env`, `customer_profile`, behavior injection).
2. Add `headers` to `BehaviorConfig`; the behavior runner now merges them under auth headers (`behavior.headers` overrides shared `target.headers`).
3. Harden `auth_runtime._normalize_headers` to drop `"None"`/`""` header values as a final safety net.

## Tests (all fail on pre-fix tree, verified in a worktree)

- `TestUnsetEnvVarNoneFiltering` — full env-expansion + flatten pipeline (6 tests)
- `test_resolve_auth_runtime_drops_none_string_*` — 2 tests
- `test_behavior_runner_merges_config_headers_into_client` — fails pre-fix with `KeyError: 'X-Tenant-Id'`
- `test_shared_headers_reach_resolved_behavior_config` — fails pre-fix with `AttributeError: 'BehaviorConfig' object has no attribute 'headers'`

Full suite: `2031 passed, 34 skipped`. ruff + mypy clean.

Fixes #293

### DIFF
--- a/nuguard.yaml.example
+++ b/nuguard.yaml.example
@@ -115,6 +115,11 @@ behavior:
   # Canary seed file — same file used by redteam canary scanning.
   # canary: ./canary.json
 
+  # Extra HTTP headers added to every request (overrides the shared
+  # target.headers block for behavior only).
+  # headers:
+  #   X-Tenant-Id: "tenant-1"
+
   # Per-request HTTP timeout in seconds (default: 60)
   # Increase for slow LLM-backed pipelines.
   request_timeout: 60
@@ -179,7 +184,8 @@ behavior:
 
 # ─── Red-team mode ────────────────────────────────────────────────────────────
 # nuguard redteam exercises adversarial scenarios to find break points.
-# target.url, target.auth, target.endpoint are inherited from the target: block above.
+# target.url, target.auth, target.endpoint, target.headers are inherited from
+# the target: block above.
 # Override any of them here when redteam needs different credentials than behavior.
 redteam:
 
--- a/nuguard/behavior/runner.py
+++ b/nuguard/behavior/runner.py
@@ -813,7 +813,13 @@ class BehaviorRunner:
 
         # Merge explicit behavior.headers (shared target.headers or behavior.headers)
         # underneath bootstrapped/auth headers so auth always wins on conflicts.
-        _cfg_headers: dict[str, str] = dict(getattr(self._config, "headers", None) or {})
+        # Normalize first so a literal "None"/"" string header value (unset
+        # ${VAR} = None -> stringified anywhere upstream) can't reach the target.
+        from nuguard.common.auth_runtime import _normalize_headers  # noqa: PLC0415
+
+        _cfg_headers: dict[str, str] = _normalize_headers(
+            getattr(self._config, "headers", None)
+        )
         if _cfg_headers:
             bootstrap_headers = {**_cfg_headers, **(bootstrap_headers or {})}
 
--- a/nuguard/behavior/runner.py
+++ b/nuguard/behavior/runner.py
@@ -811,6 +811,12 @@ class BehaviorRunner:
             _log.debug("_build_client: bootstrap skipped: %s", exc)
             bootstrap_headers = getattr(runtime, "initial_headers", {}) or {}
 
+        # Merge explicit behavior.headers (shared target.headers or behavior.headers)
+        # underneath bootstrapped/auth headers so auth always wins on conflicts.
+        _cfg_headers: dict[str, str] = dict(getattr(self._config, "headers", None) or {})
+        if _cfg_headers:
+            bootstrap_headers = {**_cfg_headers, **(bootstrap_headers or {})}
+
         # Merge login-response identity/session fields and SBOM context hints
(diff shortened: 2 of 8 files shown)


KanishkThamman (commented):
Reviewed. Core fix is correct and well-tested — verified full suite (2031 passed, 34 skipped, matches description exactly), ruff/mypy clean on all 3 changed files.

Minor: the description frames `_normalize_headers`'s new `"None"`/`""`-string hardening as a general "final safety net" for header attachment, but it isn't actually wired into either request path that attaches headers — `behavior/runner.py:813-818` and `redteam/orchestrator.py:807-810` both merge headers via plain dict unpacking, bypassing `_normalize_headers` entirely. Not dangerous today since the primary bug (unset env var → `None`) is already caught upstream in `config.py`'s filters, but the residual gap means a header explicitly set to the literal string `"None"` would still flow through unfiltered in both paths. Worth a narrow follow-up wiring `_normalize_headers` into those two merge points.

Safe to merge as-is.

nikhilpatidar:
@KanishkThamman Implemented the suggested narrow follow-up in commit 8423a8c8 on fix/config-none-header-leak: `_normalize_headers` is now wired into both request-path merge points that previously bypassed it:

- `behavior/runner.py` (`_build_client`): the explicit `behavior.headers` / shared `target.headers` merge is normalized before being layered under bootstrapped/auth headers.
- `redteam/executor/orchestrator.py` (bootstrap): `_extra_headers` → `effective_headers` is normalized before the bootstrapped session headers override it.

So a header explicitly set to the literal string `None` (or ``) can no longer flow through to the target in either path — the safety net now covers attachment, not just auth-seeding. Verified: tests/behavior/ + tests/redteam/ (1059 passed, 5 skipped), ruff + mypy clean.

KanishkThamman (approved):

PR #296, opened 2026-08-14
fix(config): unset ${VAR} in scalar redteam fields no longer crashes load_config

### LINKED ISSUE #295: config: unset ${VAR} in scalar redteam fields crashes nuguard redteam with a raw ValidationError
## Summary

An unset `${VAR}` in a **scalar** redteam field crashes config loading with a raw pydantic `ValidationError` — and `nuguard redteam` does not wrap `load_config` in a try/except (unlike `nuguard behavior`), so the traceback escapes to the user.

## Root cause

`_expand_env_vars` turns an unset `${VAR}` into `None`, but `_flatten_yaml` stringifies it into the literal `"None"` for scalar fields (`str(redteam["engine"])` → `"None"`). The `Literal`-typed model fields then reject it:

- `redteam.engine: ${RT_ENGINE}` (unset) → `redteam_engine: Literal["v1","v2"]` rejects `'None'` → **crash**
- `redteam.guided_mutation_mode: ${GMM}` (unset) → same crash
- `redteam.scenarios: [${M}, "policy-violation"]` (unset) → `[None, "policy-violation"]` → `list[str]` rejects `None` → **crash**

Path-valued fields silently carry a literal `'None'` path into runtime:
- `redteam.prompt_cache_dir: ${PCD}` (unset) → `'.../None'` → `PromptCache.save` **creates a directory literally named `None`**
- `redteam.catalog_path`, `redteam.emit_pytest_dir` → `Path('.../None')`

## Fix (PR to follow)

Skip the flatten key when the expanded value is `None` (matching the established `credentials`/LLM-block pattern), so the field falls back to its model default; filter `None` items from the `scenarios` list.

## Tests

`TestUnsetEnvVarScalarFields` (7 tests) — all but the positive case fail on the pre-fix tree, where `load_config` raises `ValidationError`.

nikhilpatidar:
## Problem

An unset `${VAR}` in a **scalar** redteam field crashes config loading with a raw pydantic `ValidationError`. `nuguard redteam` does not wrap `load_config` in try/except (unlike `nuguard behavior`), so the traceback escapes to the user.

`_expand_env_vars` turns an unset `${VAR}` into `None`, but `_flatten_yaml` stringifies it into the literal `"None"` for scalar fields:

- `redteam.engine: ${RT_ENGINE}` (unset) → `Literal["v1","v2"]` rejects `'None'` → **crash**
- `redteam.guided_mutation_mode: ${GMM}` (unset) → same crash
- `redteam.scenarios: [${M}, "policy-violation"]` → `list[str]` rejects `None` → **crash**
- Path fields (`prompt_cache_dir`, `catalog_path`, `emit_pytest_dir`) silently carry `Path('.../None')` → `PromptCache.save` creates a directory literally named `None`

## Fix

Skip the flatten key when the expanded value is `None` (matching the established `credentials`/LLM-block pattern) so the field falls back to its model default; filter `None` items from the `scenarios` list.

## Tests

`TestUnsetEnvVarScalarFields` (7 tests). On the pre-fix tree all 6 negative cases fail (verified in a worktree: `load_config` raises `ValidationError`); with the fix, defaults are restored and no crashes.

Full suite: `2038 passed, 34 skipped`. ruff + mypy clean.

Fixes #295

### DIFF
--- a/nuguard.yaml.example
+++ b/nuguard.yaml.example
@@ -115,6 +115,11 @@ behavior:
   # Canary seed file — same file used by redteam canary scanning.
   # canary: ./canary.json
 
+  # Extra HTTP headers added to every request (overrides the shared
+  # target.headers block for behavior only).
+  # headers:
+  #   X-Tenant-Id: "tenant-1"
+
   # Per-request HTTP timeout in seconds (default: 60)
   # Increase for slow LLM-backed pipelines.
   request_timeout: 60
@@ -179,7 +184,8 @@ behavior:
 
 # ─── Red-team mode ────────────────────────────────────────────────────────────
 # nuguard redteam exercises adversarial scenarios to find break points.
-# target.url, target.auth, target.endpoint are inherited from the target: block above.
+# target.url, target.auth, target.endpoint, target.headers are inherited from
+# the target: block above.
 # Override any of them here when redteam needs different credentials than behavior.
 redteam:
 
--- a/nuguard/behavior/runner.py
+++ b/nuguard/behavior/runner.py
@@ -811,6 +811,12 @@ class BehaviorRunner:
             _log.debug("_build_client: bootstrap skipped: %s", exc)
             bootstrap_headers = getattr(runtime, "initial_headers", {}) or {}
 
+        # Merge explicit behavior.headers (shared target.headers or behavior.headers)
+        # underneath bootstrapped/auth headers so auth always wins on conflicts.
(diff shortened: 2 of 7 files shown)


KanishkThamman:
Closing in favor of #294, which fixes the same unset-${VAR}-leaks-'None' bug more completely (covers headers/app_env/customer_profile consistently) without the undisclosed scope creep in this PR.

PR #303, opened 2026-08-17
fix(validate): build agent_routing scenarios from component_type

### LINKED ISSUE #302: [Bug]: validate --workflows agent_routing silently no-ops (reads nonexistent node_type)
## Expected Behavior

`nuguard validate --workflows agent_routing` should build one AGENT_ROUTING scenario per SBOM AGENT node and execute them, per the workflow contract (`ValidateScenarioType.AGENT_ROUTING`, `nuguard/models/validate.py`).

## Current Behavior

The workflow silently no-ops even when the SBOM contains AGENT nodes:

- `nuguard/validate/scenarios.py:66` filters nodes with `getattr(node, "node_type", None) == "AGENT"` — but `Node` has **no** `node_type` attribute; its field is `component_type: ComponentType` (`nuguard/sbom/models.py:741`). So `agent_names` is always `[]`, and the code logs a misleading warning: "no AGENT nodes found in SBOM. Skipping."
- `nuguard/validate/runner.py:248` has the same stale `node_type` read when initialising AGENT capability entries, so capability reporting also omits agents.

Verified on current `main`: `_agent_routing_scenarios_from_sbom()` returns `[]` for an SBOM containing an AGENT node.

## Steps to Reproduce

1. Build an `AiSbomDocument` with a `ComponentType.AGENT` node.
2. Call `build_scenarios(ValidateConfig(workflows=["agent_routing"]), sbom=doc)`.
3. Observed: zero `AGENT_ROUTING` scenarios; a warning claims no AGENT nodes exist.

## Possible Solution

Compare against `node.component_type` (the actual field, a `ComponentType` enum) instead of the nonexistent `node_type` attribute, in both `scenarios.py` and `runner.py`.

## Context

The documented agent-routing validation never runs — a silent false negative for agent-based apps.

## Environment

- Python version: 3.13
- OS: Linux
- Relevant config: `validate.workflows: [agent_routing]`

Abhipsita-coder:
## PR Type

- [x] Bug fix
- [ ] Feature

## What

Fixes `nuguard validate --workflows agent_routing` silently no-oping.

- `nuguard/validate/scenarios.py` — `_agent_routing_scenarios_from_sbom` now filters SBOM nodes by `node.component_type == ComponentType.AGENT` instead of the nonexistent `node_type` attribute.
- `nuguard/validate/runner.py` — the AGENT capability-entry initialisation uses the same corrected comparison.

## Why

Reported in issue #302. `Node` (`nuguard/sbom/models.py`) has a `component_type: ComponentType` field, **no** `node_type` attribute. The old `getattr(node, "node_type", None) == "AGENT"` filter always matched zero nodes, so:

- `agent_routing` workflows produced **zero** scenarios and logged a misleading `"workflow 'agent_routing' requested but no AGENT nodes found in SBOM. Skipping."` warning — even when the SBOM contained AGENT nodes.
- `ValidateRunner` never seeded AGENT capability entries, so capability reporting omitted agents.

This is a silent false negative for agent-based applications: the documented agent-routing validation never ran.

## Root Cause

Stale attribute name: the consumer code was written against a `node_type` field that does not exist on the `Node` model. Both call sites were affected.

## How

Compare against the actual field — `node.component_type == ComponentType.AGENT` (`ComponentType` is a `str` enum with `AGENT = "AGENT"`, `nuguard/sbom/types.py`). Imported at top level in both modules.

## Test Steps

- [x] Reproduce on `main`: build an SBOM with AGENT nodes, call `build_scenarios(ValidateConfig(workflows=["agent_routing"]), sbom=...)` → zero scenarios + misleading warning.
- [x] Confirm on this branch: exactly one `AGENT_ROUTING` scenario per AGENT node; TOOL/MODEL nodes excluded.
- [x] Two regression tests added to `nuguard/validate/tests/test_scenarios.py` (both fail pre-fix, pass post-fix — verified in a temporary worktree).

## Checks

- [x] `make test` passes — focused suites green (`nuguard/validate/tests/` + `tests/cli/test_validate_command.py`: 35 passed)
- [x] `make lint` passes (`ruff check` + `mypy`) — clean
- [x] No unrelated reformatting: only the 3 logical hunks + new tests
- [x] Added/updated tests covering this change (regression test for the bug)

Fixes #302

### DIFF
--- a/nuguard/validate/runner.py
+++ b/nuguard/validate/runner.py
@@ -25,6 +25,7 @@ from nuguard.models.validate import (
 from nuguard.redteam.target.canary import CanaryConfig, CanaryScanner
 from nuguard.redteam.target.client import TargetAppClient
 from nuguard.redteam.target.session import AttackSession
+from nuguard.sbom.types import ComponentType
 from nuguard.validate.scenarios import build_scenarios
 
 if TYPE_CHECKING:
@@ -247,8 +248,8 @@ class ValidateRunner:
             try:
                 for node in self._sbom.nodes:
                     if (
-                        getattr(node, "node_type", None) == "AGENT"
-                        and getattr(node, "name", None)
+                        node.component_type == ComponentType.AGENT
+                        and node.name
                         and node.name not in cap_entries
                     ):
                         cap_entries[node.name] = CapabilityEntry(
--- a/nuguard/validate/scenarios.py
+++ b/nuguard/validate/scenarios.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 from nuguard.common.logging import get_logger
 from nuguard.config import ValidateConfig
 from nuguard.models.validate import ValidateScenario, ValidateScenarioType
+from nuguard.sbom.types import ComponentType
 
 if TYPE_CHECKING:
     from nuguard.sbom.models import AiSbomDocument
@@ -63,7 +64,7 @@ def _agent_routing_scenarios_from_sbom(sbom: "AiSbomDocument") -> list[ValidateS
         agent_names = [
             node.name
             for node in sbom.nodes
-            if getattr(node, "node_type", None) == "AGENT" and getattr(node, "name", None)
+            if node.component_type == ComponentType.AGENT and node.name
         ]
     except Exception:
         agent_names = []
--- a/nuguard/validate/tests/test_scenarios.py
+++ b/nuguard/validate/tests/test_scenarios.py
@@ -120,3 +120,53 @@ def test_boundary_assertion_forbid_pattern():
     )
     scenarios = build_scenarios(cfg)
     assert scenarios[0].forbid_pattern == r"\d{3}-\d{2}-\d{4}"
+
+
+def test_agent_routing_builds_one_scenario_per_agent_node():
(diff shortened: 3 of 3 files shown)


KanishkThamman (commented):
Reviewed. Fixes a real dead-comparison bug: `node.node_type` doesn't exist on `Node` (always `None`), so the `== "AGENT"` check at both call sites was always false. Correctly replaced with `node.component_type == ComponentType.AGENT`, the actual field. Verified by reverting just the fix hunks — both new tests failed exactly as described (0==1 / 0==2 with the misleading warning), confirming these are genuine regression tests, not tautological ones.

Confirmed no other latent instance of this bug exists elsewhere in the codebase (grepped all `node_type` usages — everything else either uses the correct local helpers or is an unrelated field on a different model). Tests: 12/12 and 35/35 pass (matches claimed count), ruff/mypy clean.

Nits (non-blocking): the now-redundant `and node.name` truthy check could be dropped since `name` is a required non-optional field, but it's harmless. Also, `runner.py`'s identical fix (capability-entry seeding) only got indirect coverage via mypy/ruff — no dedicated unit test at that specific call site, only `scenarios.py` got one.

Safe to merge as-is.

KanishkThamman (approved):

PR #333, opened 2026-08-21
fix(policy): build the LLM client from config in policy check --llm

### LINKED ISSUE #325: policy check --llm ignores configured LLM model, falls back to Gemini default
## Description

Running \`nuguard policy check --llm\` (compliance framework assessment path) uses the hardcoded default Gemini model instead of the model configured in \`nuguard.yaml\` (\`llm.model\`).

## Root cause

In \`nuguard/cli/commands/policy.py\`, the \`check\` command's compliance-assessment branch calls:

\`\`\`python
run_compliance_assessment(doc, framework=fw_name, enable_llm=enable_llm)
\`\`\`

without passing an \`llm\` client. In \`nuguard/policy/assessment.py\`, when \`llm=None\` and a control needs LLM fallback, it calls \`llm_synthesize(control, snapshot, llm=None)\`. In \`nuguard/policy/synthesizer.py\`, that falls back to constructing a bare \`LLMClient()\` with no model override, which resolves to \`_DEFAULT_MODEL = \"gemini/gemini-3.1-flash-lite\"\` in \`nuguard/common/llm_client.py\`, ignoring \`cfg.litellm_model\`/\`cfg.litellm_api_key\`/\`cfg.litellm_api_base\`.

By contrast, \`policy compile --llm\` (same file) correctly builds \`LLMClient(model=cfg.litellm_model, api_key=cfg.litellm_api_key, api_base=...)\` and passes it through.

## Fix

Build the \`LLMClient\` from config in the \`check\` command the same way \`compile_policy\` does, and pass it to \`run_compliance_assessment(..., llm=llm_client)\`.

Abhipsita-coder:
## Summary

\
uguard policy check --llm\ (compliance-framework assessment path) ignored \llm.model\ from \
uguard.yaml\: it called \
un_compliance_assessment(...)\ without an \llm\ client, so \llm_synthesize\ fell back to a bare \LLMClient()\ and silently used the hardcoded Gemini default.

## Fix

Build the \LLMClient\ from config in the \check\ command — the same pattern \policy compile --llm\ already uses — and pass it through via \
un_compliance_assessment(..., llm=llm_client)\.

## Testing

New CLI regression test (\	est_check_llm_builds_client_from_config\) asserts the assessment receives a client constructed with the configured model and API key. All 20 tests in \	ests/cli/test_policy_cli.py\ pass; ruff + mypy clean.

Fixes #325

### DIFF
--- a/nuguard/cli/commands/policy.py
+++ b/nuguard/cli/commands/policy.py
@@ -363,12 +363,24 @@ def check(
         fw_name = framework or "custom"
         _console.print(f"\n[bold]Running compliance assessment:[/bold] {fw_name} …")
 
+        llm_client = None
+        if enable_llm:
+            from nuguard.common.llm_client import LLMClient
+
+            _model = cfg.litellm_model or ""
+            llm_client = LLMClient(
+                model=cfg.litellm_model,
+                api_key=cfg.litellm_api_key,
+                api_base=cfg.litellm_api_base if _model.startswith("azure") else None,
+            )
+
         try:
             assessment = asyncio.run(
                 run_compliance_assessment(
                     doc,
                     framework=fw_name,
                     enable_llm=enable_llm,
+                    llm=llm_client,
                 )
             )
         except Exception as exc:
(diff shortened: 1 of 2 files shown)


ganesh-nuguard (approved):
Looks good
Thanks for taking care

PR #338, opened 2026-08-21
fix(sbom): link fluent C# call chains in find_calls

Abhipsita-coder:
## Summary

Part 2 of the #260 hardening series: fluent invocation chains in `find_calls`.

- `find_calls` previously recorded only the identifier chain immediately before the current `(`, so in `Trainer().Fit(...)` the `Fit` call lost both its receiver and the final assignment target.
- All call expressions are now parsed first; each call joined to its predecessor by `.` or `?.` inherits the **full preceding invocation expression** as its receiver and, when it has no assignment of its own, the chain's assignment target.
- The optional *names* filter is applied **after** linking, so chains resolve even when intermediate calls are filtered out (`find_calls(src, {"Fit"})` still sees `mlContext.Transforms.Text("x").Fit(d)`).
- ML.NET: chained `Fit` calls now carry `.Transforms`/`.Trainers` receivers, so `Fit` is routed to its own trained-model handler instead of being mis-emitted as a pipeline stage.

## Testing

3 new regression tests (receiver/assignment inheritance, linking across name filters, end-to-end ML.NET trained-model emission from a fluent chain). Full sbom test suite green except two pre-existing unrelated csproj failures on main.

Part of #260

### DIFF
--- a/nuguard/sbom/adapters/csharp/_source.py
+++ b/nuguard/sbom/adapters/csharp/_source.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import re
 from bisect import bisect_right
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from typing import Iterable
 
 from ...core.csharp_parser import CSharpMethodDeclaration, CSharpParseResult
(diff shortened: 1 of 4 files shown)


Abhipsita-coder:
Closing as superseded by #379.

This PR was based on stale main (pre-chaining fork base). Auditing it against current upstream/main showed that fluent chain linking (receivers, assignment-target inheritance, names-filter-after-linking) and the ML.NET `Fit` routing are all already present in main — its own chain-linking and ML.NET tests pass verbatim on current main.

The only genuinely missing behavior was that reconstructing a chained constructor receiver dropped the `new` keyword (`new MyTrainer().Fit(...)` yielded receiver `"MyTrainer()"` instead of `"new MyTrainer()"`).

That single behavior is now isolated and fixed in #379, which is based on current main and contains only the focused production change (preserve the `new` prefix on chained constructor receivers) plus its regression test. Nothing else from this PR was carried over.

PR #339, opened 2026-08-21
fix(sbom): harden C# Semantic Kernel model and plugin registration edge cases

Abhipsita-coder:
## Summary

Part 3 of the #260 hardening series: Semantic Kernel registration edge cases.

- **No fabricated model nodes** - `AddOpenAIChatCompletion(modelId)` with an unresolved expression used to derive a model name from the registration method itself (producing MODEL nodes named `OpenAI`/`AzureOpenAI`). The registration is now skipped when no model/deployment identifier can be resolved.
- **Registration-specific plugin names** - plugin extraction always assumed positional argument zero is the plugin name. `AddFromObject` / `ImportPluginFromObject` and `AddFromPromptDirectory` / `ImportPluginFromPromptDirectory` place the name in argument one; those overloads now resolve it correctly.

## Testing

2 new regression tests: unresolved model registration emits no MODEL node and no method-derived display name; plugin names resolve per overload (`MyPlugin`, `custom-name`, `prompt-plugin`). Full sbom suite green except two pre-existing unrelated csproj failures on main.

Part of #260

### DIFF
--- a/nuguard/sbom/adapters/csharp/_source.py
+++ b/nuguard/sbom/adapters/csharp/_source.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import re
 from bisect import bisect_right
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from typing import Iterable
 
 from ...core.csharp_parser import CSharpMethodDeclaration, CSharpParseResult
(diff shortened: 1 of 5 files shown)


Abhipsita-coder:
Closing as superseded: current upstream/main already covers both behaviors this PR added. `if not model_name: continue` (skip unresolved model/deployment registrations instead of fabricating a MODEL node from the method name) is present, and plugin-name resolution uses the richer `_PLUGIN_NAME_POSITIONS` mapping (positional arg 1 and 2, named `pluginName`/`name` first) — both of this PR's regression tests pass verbatim against current main. No unique production behavior remains (coverage audit vs current main, 2026-08-26).

PR #332, opened 2026-08-21
MAIN - fix(config): fall back to model defaults for scalar redteam fields with unset env vars

Abhipsita-coder:
## Summary

An unset \\\ in a **scalar** redteam field crashed config loading: \_expand_env_vars\ resolves it to \None\, but \_flatten_yaml\ stringified it into the literal \'None'\, which

- Literal-typed fields (\
edteam_guided_mutation_mode\) reject with a raw pydantic \ValidationError\ — the traceback escapes to the user since \
uguard redteam\ does not wrap \load_config\,
- path-valued fields (\
edteam.catalog_path\, \
edteam.prompt_cache_dir\, \
edteam.emit_pytest_dir\) carried into runtime as a directory literally named \None\,
- list fields (\
edteam.scenarios\) rejected the embedded \None\ item.

## Fix

- Drop top-level \None\ values from the \
edteam:\ section in \_flatten_yaml\, matching the established credentials/LLM-block pattern, so affected fields fall back to their model defaults.
- Filter \None\ entries from the \scenarios\ list so remaining explicit goals still run.

## Testing

New \TestUnsetEnvVarScalarFields\ (7 tests): guided_mutation_mode default fallback, set-var still applies, scenario list filtering (partial and full), and the three path fields. All 56 tests in \	ests/test_config.py\ pass; ruff + mypy clean.

Fixes #295

PR #360, opened 2026-08-24
fix(config): fall back to model defaults for scalar redteam fields with unset env vars

### LINKED ISSUE #295: config: unset ${VAR} in scalar redteam fields crashes nuguard redteam with a raw ValidationError

rangoel-nu:
## Summary

Re-applies #332's fix onto `develop` (that PR targets `main`, which has drifted
significantly from `develop`; cherry-picked here as a clean, verified apply).

An unset `${VAR}` in a scalar redteam field was stringified to the literal
`'None'` by `_flatten_yaml`, so `Literal`-typed fields (e.g.
`redteam.guided_mutation_mode`) rejected it with a raw `ValidationError`, and
path-valued fields (`catalog_path`, `prompt_cache_dir`, `emit_pytest_dir`)
carried the literal string `'None'` into runtime — e.g. `PromptCache.save`
would create a directory literally named `None`.

Fix: drop top-level `None` values from the `redteam` section before any
per-field handling runs (matching the established `credentials`/LLM-block
pattern), and filter `None` entries out of the `scenarios` list.

Fixes #295

## Testing

- Reproduced the crash against current `develop` before applying the fix
  (`redteam.guided_mutation_mode: ${GMM}` with `GMM` unset → raw
  `pydantic.ValidationError`).
- Confirmed the fix resolves it — falls back to the model default instead.
- `uv run ruff check nuguard/config.py` — clean.
- `uv run mypy nuguard/config.py` — clean.
- `uv run pytest tests/test_config.py` — 61 passed (includes the 7 new
  `TestUnsetEnvVarScalarFields` cases from #332).
- Full suite: `uv run pytest nuguard/ tests/` — 4065 passed, 71 skipped, 0
  failures.

### DIFF
--- a/nuguard/config.py
+++ b/nuguard/config.py
@@ -251,6 +251,10 @@ def _flatten_yaml(data: dict[str, Any]) -> dict[str, Any]:
                 flat["redteam_auth_login_flow"] = _shared_auth["login_flow"]
 
     # Redteam section — overrides shared target block when keys are present.
+    # Drop keys whose env-var interpolation resolved to None (unset ${VAR}
+    # without a ``:-default``) so scalar fields keep their model defaults
+    # instead of crashing validation as the literal string 'None' — matching
+    # the established credentials/LLM-block handling below.
     # Both `endpoint:` and `target_endpoint:` are accepted as aliases, mirroring
     # the shared ``target:`` block (which accepts both names). The shared block
     # already does this on lines 199-203; the override must accept the same
@@ -258,7 +262,7 @@ def _flatten_yaml(data: dict[str, Any]) -> dict[str, Any]:
     # in nuguard.yaml.example line 47) see their override take effect.
     # Precedence matches the shared block: ``endpoint`` wins over
     # ``target_endpoint`` when both are set in the same block.
-    redteam = data.get("redteam", {}) or {}
+    redteam = {k: v for k, v in (data.get("redteam", {}) or {}).items() if v is not None}
     if "target" in redteam:
         flat["target_url"] = redteam["target"]
     if "endpoint" in redteam:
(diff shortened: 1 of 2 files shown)


rangoel-nu:
Recreating with correct branch-name convention (cherry-pick/* for a develop-targeted PR, per check-base).

PR #391, opened 2026-08-27
feat(probe): OpenAPI schema-driven nested payload detection

### LINKED ISSUE #386: [Bug]: Improve API Detection for Chat & Login Endpoints
### From the original gap list (still open)

**OpenAPI/Swagger schema not fetched**
If the target exposes `/openapi.json`, `/swagger.json`, or `/docs`, the exact request schema (required fields, field types) and response schema are available for free. The pipeline never attempts to fetch this. Doing so before the probe would replace 8 blind payload guesses with exact knowledge of what the endpoint expects.

**422 validation error body not parsed**
FastAPI/Pydantic returns `{"detail": [{"loc": ["body", "message"], "msg": "field required"}]}` on validation failure. The probe currently hits a 422 and moves to the next shape without reading the body. Parsing this response would immediately reveal the correct field name without trying additional shapes.

**OpenAI-compatible `messages` format missing from probe shapes**
`_PROBE_PAYLOADS` has 8 shapes (`message, phrases, input, query, prompt, text, content, msg`) but no `messages` entry. Any LiteLLM wrapper, OpenAI-compatible proxy, or Anthropic Messages API endpoint will 422 on all 8 shapes and be recorded with the wrong `payload_key="message"`. This is the single most widely deployed chat format today.

**Streaming responses silently skipped**
When an endpoint streams its response (SSE / NDJSON / chunked), `resp.json()` throws and the endpoint is silently skipped. Many modern agent APIs (`/stream`, `/chat/stream`, LangGraph streaming endpoints) are completely invisible to the probe.

**WebSocket endpoints unsupported**
Some LangGraph Cloud and OpenAI Realtime API wrappers use WebSocket for chat. No probe path handles `ws://` or `wss://` URIs at all.

**5xx exits the payload shape loop with only the first shape saved**
```python
if status >= 500:
    if server_error_fallback is None:
        server_error_fallback = (path, pay_key, pay_list)
    break  # moves to next path immediately
```
A path that 5xx's on `message` but would 2xx on `messages` is saved as `(path, "message", False)` and never retried with other shapes. The correct payload key for that path is permanently missed.

**Response key heuristic too narrow**
`_looks_like_chat_response` accepts only: `response, content, prognosis, text, output, answer, result, reply, choices, messages`. Endpoints returning `bot_response`, `assistant_message`, `generated_text`, `completion`, or nested structures like `data.message` all fall through to raw-text parsing, which can fail and cause the endpoint to be skipped entirely.

**Auth token auto-detection still only covers top-level flat keys**
The SBOM-based `basic→login_flow` upgrade in `resolve_auth_config_with_sbom_fallback` auto-scans for `token, access_token, jwt, id_token, auth_token` — all top-level flat keys. The `auth_token_response_key` override added in `fc8da64d` is a manual escape hatch. A target with `data.token` or `result.access_token` still needs the caller to know the nesting path in advance; the auto-upgrade won't find it.

**Login endpoint regex misses non-standard paths**
`resolve_auth_config_with_sbom_fallback` only matches `/login|/auth|/signin|/token|/authenticate`. Paths like `/api/v1/users/session`, `/connect`, `/api/auth/local` (Strapi), `/me/token` are never recognized as login endpoints — credentials are sent as raw Basic auth to the chat endpoint instead.

**Session UUID auto-generation causes empty agent responses**
When a `session_id` or `conversation_id` is required in the payload and no value can be sourced from the login response or config, `apply_sbom_context_hints` generates a random UUID. The agent doesn't recognize the session and returns empty/error responses, and the discovery conversation extracts nothing useful.

**No GraphQL probing**
GraphQL APIs use `POST /graphql` with `{"query": "{ chat(message: \"...\") { reply } }"}`. The probe sends flat REST-style JSON and always gets a 400. The `/graphql` path isn't in the fallback list.

**Domain detection for discovery messages is implicit**
The domain (banking/airline/healthcare/generic) controlling which discovery messages are sent is inferred from SBOM framework tags, not from the live agent's actual self-description. If the SBOM summary doesn't carry the right tags, generic messages are sent, producing lower-quality golden data.

**Probe timeout vs. bootstrap timeout mismatch**
Bootstrap uses 30s per request (cold-start friendly). `probe_chat_endpoints` defaults to 15s. A serverless target that cold-starts in 20s will pass bootstrap but the probe will time out and fall back to a wrong endpoint.

**No multi-turn probing**
The probe sends one message and inspects the first response. Some agents require an initial greeting exchange before producing structured output. A two-turn probe would be more reliable for detecting the correct payload format.

---

### New findings from code analysis

**`auth_username` not forwarded to `apply_sbom_context_hints` — identity-candidate rotation is dead code**
`session_resolver.py` calls `apply_sbom_context_hints(sbom, chat_path, merged_extras, login_extras)` without passing `auth_username` (defaults to `None`). The identity-candidate rotation feature in `discovery.py` — which derives username variants (full → strip domain → first name) and retries the discovery conversation with each — is therefore unreachable through the standard `resolve_target_session` pipeline. Any endpoint requiring `user_id` in the body with no login-response value will always get a random UUID, never a username-derived candidate.

**Two independent probe calls against the same target — prescan 15s cap can abort the run**
`run_prescan_discovery` in `executor_utils.py` issues its own `probe_chat_endpoints` call with `timeout=min(timeout, 15.0)` and raises a hard `RuntimeError` if no endpoint is found. The main run then calls `resolve_target_session` which probes independently. For a serverless target with a 20s cold-start, the prescan probe times out at 15s and the run is aborted before the main probe (which could use a longer timeout) ever fires. This makes gap #13 more impactful than described — it doesn't just affect endpoint quality, it causes hard run failures.

**Multi-step conversation flows — new gap class entirely**
Commit `fc8da64d` explicitly notes: *"the real chat endpoint requires a `POST /api/conversations` to create a session ID, then `POST /api/conversations/{conversation_id}/messages` to send a message — nothing in the platform or the `nuguard` package currently supports this."* The probe sends a one-shot POST and has no way to chain requests or carry a server-generated ID as a path parameter. Multi-tenant chat platforms and many commercial agent SDKs use this create-then-send pattern.

**`run_with_endpoint_fallback` (`91e163b7`) cycles known candidates but doesn't widen discovery**
`run_with_endpoint_fallback` retries public API calls across pre-discovered SBOM endpoint candidates on `endpoint_rejected` or `target_unavailable` failures. This improves reliability for already-known paths, but it doesn't widen the discovery net — it only rotates through what `discover_chat_candidates_from_sbom` already returned. Targets with no SBOM `API_ENDPOINT` nodes still get a single attempt against the guessed path.

ganesh-nuguard:
Closes #386

## Summary

Adds automatic detection of nested JSON payload structures from OpenAPI/Swagger schemas, so nuguard can probe and red-team targets that expect structured message objects (e.g. `{"message": {"role": "user", "content": "..."}}`) instead of plain strings — without any manual `chat_payload_key` configuration.

### What changed

**`endpoint_probe.py` — `ProbeResult` dataclass**
- Replaces the raw `tuple[str, str, bool]` return with a `ProbeResult` dataclass that adds an optional `value_template` field.
- `__iter__` yields only the 3 positional fields so all existing callers that do `path, key, is_list = result` are unchanged.

**`endpoint_probe.py` — `_build_object_template()`**
- Introspects any OpenAPI object schema to build a field-name-agnostic payload template.
- First pass places `_CHAT_TEXT_SENTINEL` on the most likely content field (`content` > `text` > `message` > `body` > `input` > `prompt` > `query` > `msg`).
- Second pass fills required fields with type-appropriate defaults (`""`, `0`, `False`, first enum value, nested object up to depth 2).
- Returns `None` when no recognisable content field is found → falls back to plain string (no regression).

**`client.py` — `_build_chat_payload_value()`**
- New first branch: when `value_template` is set, replaces any field whose value equals `_CHAT_TEXT_SENTINEL` with the actual chat text at send time. Works for any field name — not hardcoded to `content`.

**`orchestrator.py` — `_maybe_probe_endpoints()`**
- Added Option B: when the endpoint is already known but the key is still the default `"message"`, fires a `hint_path` probe against that specific endpoint to detect the nested shape via OpenAPI.
- Fixes pre-flight 422 failures on FastAPI/Pydantic targets that require a structured message object.
- Call site updated: `_maybe_probe_endpoints` now runs when `not chat_path OR chat_payload_key == "message"` (was: only when `not chat_path`).

**Propagation chain**
`ProbeResult.value_template` → `TargetSessionConfig.chat_payload_value_template` → `build_target_app_client(chat_payload_value_template=...)` → `TargetAppClient._chat_payload_value_template`

### Validation

Tested end-to-end against the NuGuard dev chatbot (`https://api-dev.nuguard.ai/api/chat`):
- OpenAPI schema fetched from `/openapi.json` ✓
- Payload key `messages` (list=True) auto-detected ✓
- Bootstrap + pre-flight passed ✓
- 131 scenarios published, ~80 executed before token expiry ✓
- 2 findings produced (context flooding data exfiltration, system prompt disclosure) ✓

Unit tests: `pytest tests/common/test_endpoint_probe.py nuguard/common/tests/ nuguard/redteam/tests/test_message_history_payload.py` — 61 passed.

### DIFF
--- a/nuguard/common/endpoint_probe.py
+++ b/nuguard/common/endpoint_probe.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 import json
 import re
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING
 
 import httpx
@@ -25,6 +26,31 @@ if TYPE_CHECKING:
     from nuguard.sbom.models import AiSbomDocument
 
 _log = get_logger(__name__)
+
+
+@dataclass
+class ProbeResult:
+    """Result of a successful chat endpoint probe.
+
+    Supports 3-value tuple unpacking (``path, key, is_list = result``) so
+    existing callers are unchanged.  Access ``.value_template`` explicitly
+    when you need the nested payload shape detected from OpenAPI schema.
+    """
+
+    path: str
+    key: str
+    is_list: bool
+    # Payload value template built from OpenAPI schema when the chat key expects
+    # a structured object (e.g. {"role": "user", "content": "..."}) rather than
+    # a plain string.  None means use a plain string (the default behaviour).
+    value_template: "dict[str, object] | None" = field(default=None, compare=False)
+
+    def __iter__(self):  # noqa: ANN204
+        # Yield only the 3 positional fields so ``a, b, c = result`` still works.
+        return iter((self.path, self.key, self.is_list))
+
+    def __len__(self) -> int:
+        return 3
 
 # Detects path-parameter placeholders — FastAPI/Express {id}, NestJS :id, or <id>.
 _HAS_PATH_PARAM_RE = re.compile(r"(:\w+|\{\w+\}|<\w+>)")
@@ -95,6 +121,16 @@ _RUNTIME_NON_CHAT_KEYS: frozenset[str] = frozenset({
 
 _CAMEL_CASE_RE = re.compile(r"(?<!^)(?=[A-Z])")
 
+# Sentinel placed in the value_template dict for the field that carries the
+# actual chat text.  Replaced at send-time regardless of the field name.
+_CHAT_TEXT_SENTINEL = "__nuguard_chat_text__"
+
+# String field names that most likely carry the chat message text, in priority
+# order.  The first match wins the sentinel in _build_object_template.
+_CONTENT_FIELD_NAMES: tuple[str, ...] = (
+    "content", "text", "message", "body", "input", "prompt", "query", "msg",
+)
+
 
 def _normalize_payload_key(key: str) -> str:
     """Normalize a payload key for comparison against _RUNTIME_NON_CHAT_KEYS.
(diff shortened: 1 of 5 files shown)


fatkobra (changes requested):
Thanks for the work on schema-driven payload detection. I’m requesting changes before merge.

**Blocking**

* `_chat_config_from_openapi()` can combine the selected endpoint with a schema from another operation. Store the winning schema with the winning path/key.
* `hint_path` does not restrict OpenAPI selection. Filter candidate operations before scoring.
* Explicit `chat_payload_key: message` can be overwritten. Explicit configuration must remain authoritative.
* Nested templates are incomplete: required child objects may be omitted, and sentinel replacement only handles top-level values. Make template creation and substitution recursive.
* A validation `422` is treated as successful verification. Do not accept a payload the server rejected as invalid.
* The template is not updated atomically with endpoint/key/list changes, which can cause missing or stale templates during pre-scan and endpoint rotation.
* Add regression tests for each case above. The reported E2E test covers the existing `messages` list shape, not the new nested-object flow.

**Non-blocking**

* Use `Refs #386` instead of `Closes #386` unless this PR resolves the full umbrella issue.
* Add a brief documentation or release-note entry for the new behavior.

Please address the blocking items and add focused tests before re-review.

PR #406, opened 2026-08-28
fix(sbom): add .mjs and .cjs to TypeScript extension dispatch

### LINKED ISSUE #416: [Bug]: SBOM extractor misses .mjs and .cjs TypeScript extensions
### Description
The SBOM extractor's TypeScript file extension dispatch only recognises `.ts` and `.tsx` files. Files with `.mjs` and `.cjs` extensions — which are valid JavaScript/TypeScript module formats used by Node.js and bundlers — silently fall through to the generic regex fallback instead of being routed to the TypeScript adapter.

### Root Cause
The extension-to-adapter dispatch table in the SBOM extractor does not include `.mjs` or `.cjs` in the TypeScript extension list. These extensions are functionally equivalent to `.js`/`.ts` for parsing purposes but are missed by the current mapping.

### Impact
Repositories using `.mjs` or `.cjs` extensions for TypeScript/JavaScript modules get degraded SBOM extraction — structural detection (framework adapters, API endpoints, models) is skipped, and only the generic regex fallback runs.

### How I Validated
- Added `.mjs` and `.cjs` to the TypeScript extension dispatch
- Added regression tests for the new extensions
- All existing tests pass

nikhilpatidar:
## PR Type

- [x] Bug fix
- [ ] Feature

## Summary

ESM (`.mjs`) and CommonJS (`.cjs`) JavaScript files were silently skipped by the tree-sitter-based TypeScript adapter pipeline, receiving only regex-based fallback detection.

## Root cause

`_TYPESCRIPT_EXTENSIONS` in `extractor/core.py` only contained `{".ts", ".tsx", ".js", ".jsx"}`. Files with `.mjs` or `.cjs` extensions — increasingly common in modern Node.js projects — were never dispatched to the TS adapter pipeline, so they missed framework-specific detection (LangChain, OpenAI Agents, LLM clients, etc.).

## Fix

Add `.mjs` and `.cjs` to `_TYPESCRIPT_EXTENSIONS` so they route through the same adapter pipeline as `.ts/.tsx/.js/.jsx`. The `_iter_ts_sources` cross-file pre-pass remains `.ts/.tsx`-only since ESM/CJS files rarely contain TypeScript-specific type exports needed for cross-file DTO resolution.

## Tests

6 new tests verifying:
- `.mjs`/`.cjs` are in the extension set
- Original extensions are preserved
- `.mjs` with OpenAI import produces MODEL detection
- `.cjs` with OpenAI `require` produces MODEL detection
- Unrelated `.mjs` produces no false positives

## Testing

```bash
uv run pytest nuguard/sbom/tests/test_mjs_cjs_dispatch.py -v
```

## Related

Addresses #386 (partial — the `.mjs`/`.cjs` dispatch gap)

Fixes #416

### DIFF
--- a/nuguard/sbom/extractor/core.py
+++ b/nuguard/sbom/extractor/core.py
@@ -110,7 +110,8 @@ _SQL_EXTENSIONS = {".sql"}
 # Jupyter notebooks: cells are extracted and parsed as Python
 _NOTEBOOK_EXTENSIONS = {".ipynb"}
 # TypeScript/JavaScript: tree-sitter (or regex fallback) via core/ts_parser
-_TYPESCRIPT_EXTENSIONS = {".ts", ".tsx", ".js", ".jsx"}
+# .mjs = ESM JavaScript, .cjs = CommonJS JavaScript — both contain JS/TS-compatible code
+_TYPESCRIPT_EXTENSIONS = {".ts", ".tsx", ".js", ".jsx", ".mjs", ".cjs"}
 # Dockerfile: extensionless file named "Dockerfile" or suffixed ".dockerfile"
 _DOCKERFILE_EXTENSIONS = {".dockerfile"}
 _DOCKERFILE_NAMES = {"dockerfile"}  # lower-cased stem match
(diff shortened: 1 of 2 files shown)


nikhilpatidar:
Superseded by #443 targeting develop (new workflow: all PRs → develop).

PR #446, opened 2026-08-29
fix(redteam): re-arm the direct-HTTP endpoint circuit breaker across passes

### LINKED ISSUE #377: redteam: direct-HTTP endpoint circuit breaker never re-arms after a transient blip
## Summary

The direct-HTTP endpoint-probe circuit breaker (introduced in #013a62f8 to isolate `TargetUnavailableError(source="endpoint_probe")` from the chat-endpoint breaker) is **latch-only**: once 3 consecutive probe failures trip it, `RedteamOrchestrator._endpoint_circuit_open` is set permanently and the orchestrator's consecutive-failure counter is never reset on a recovered response.

## Impact

A ~3-connection blip — container cold start, load-balancer re-point, transient DNS failure — silently drops **every remaining API_ATTACK / direct-HTTP scenario for the whole run**:

- `chain_status="target_unreachable"`, `0/0` turns in the report, even after the target fully recovers.
- Affects the escalation pass too: `_run_scenarios` latches the event up front via `_endpoint_circuit_open` and skips all direct-HTTP scenarios with **zero probe requests**.
- The chat breaker, by contrast, resets `consecutive_unavailable = 0` on success — only the direct-HTTP isolation breaker is sticky.

## Expected vs actual

Expected: a transient probe blip skips the rest of the pass that tripped it, then the breaker re-arms — subsequent passes (escalation) probe again, and a recovered endpoint is exercised normally. A genuinely dead endpoint stays bounded (each pass re-probes at most `_ABORT_THRESHOLD` times).

Actual: after the trip, the breaker never re-arms for the rest of the run; recovered endpoints are never probed again.

## Reproduction

See the regression tests in #PR (branch `fix/endpoint-circuit-breaker-rearm`): a two-pass `_run_scenarios` sequence against a probe that fails 3 times then recovers shows pass 2 skipping all scenarios with 0 probes pre-fix, and probing/completing them post-fix.

## Proposed fix

Scope the isolated breaker to the pass that tripped it: a fresh counter + abort event per `_run_scenarios` call (mirroring the chat breaker's per-pass counter), and drop the sticky `_endpoint_circuit_open` flag.

nikhilpatidar:
## PR Type
- [x] Bug fix
- [ ] Feature

## Root Cause
The direct-HTTP endpoint circuit breaker (`_endpoint_circuit_open`) was
latched permanently across redteam passes. When 3 consecutive
`TargetUnavailableError` exceptions tripped the breaker (e.g. during
container cold start, LB re-point, or transient DNS failure), the flag
was set to `True` and never cleared — even after the endpoint recovered.
This silently dropped every remaining `API_ATTACK` scenario (auth-bypass,
IDOR, mass-assignment, BFLA chains) for the entire run.

The chat-routed circuit breaker had the same design flaw but was masked
because the `consecutive_unavailable` counter reset per pass.

## What Changed
1. Remove the persistent `_endpoint_circuit_open` flag entirely.
2. Make the endpoint abort event **per-pass only** (fresh
   `asyncio.Event()` per `_run_scenarios` call) — a transient blip
   within a pass still aborts remaining direct-HTTP scenarios in that
   pass, but the next pass starts with a clean slate.
3. Add explanatory comments clarifying the per-pass isolation semantics.
4. Update docstrings to remove references to the removed flag.

## Why
A ~3-connection blip (common with container cold starts or k8s pod
rotation) would permanently disable all direct-HTTP attack scenarios for
the rest of the run, even after the target recovered. This defeated the
purpose of running `API_ATTACK` chains.

## How Validated
- 3 new tests in `test_endpoint_breaker_rearm.py`:
  - `test_endpoint_breaker_skips_remaining_within_pass`
  - `test_endpoint_breaker_rearms_across_passes_after_recovery`
  - `test_endpoint_breaker_rearms_across_passes_while_down`
- `uv run pytest nuguard/redteam/tests/test_endpoint_breaker_rearm.py` — all pass
- `uv run ruff check nuguard/redteam/executor/orchestrator.py` — clean

## Release Notes
The redteam direct-HTTP endpoint circuit breaker now resets between
passes. A transient outage during one pass no longer permanently disables
`API_ATTACK` scenarios for the rest of the run.

Fixes #377

### DIFF
--- a/nuguard/redteam/executor/orchestrator.py
+++ b/nuguard/redteam/executor/orchestrator.py
@@ -290,8 +290,7 @@ def _is_direct_http_only_scenario(scenario: AttackScenario) -> bool:
     steps at all (e.g. a static API_ATTACK auth-bypass/IDOR/mass-assignment/
     BFLA chain against SBOM-derived REST paths) is "direct-HTTP-only" and
     should not be caught by the chat-endpoint circuit breaker — see
-    ``TargetAppClient._consecutive_endpoint_errors`` and
-    ``RedteamOrchestrator._endpoint_circuit_open``.
+    ``TargetAppClient._consecutive_endpoint_errors``.
 
     Guided conversations always route through chat, so they are never
     direct-HTTP-only.
@@ -718,16 +717,6 @@ class RedteamOrchestrator:
         # Circuit-open flag latched when the 3-strike abort trips; consulted by
         # the escalation pass so it skips rather than re-hitting a dead target.
         self._circuit_open = False
-        # Separate circuit-open flag for direct-HTTP endpoint-probe outages
-        # (TargetUnavailableError raised with source="endpoint_probe" — see
-        # TargetAppClient._record_endpoint_error). Kept independent of
-        # self._circuit_open so an unreachable SBOM-derived REST path (e.g.
-        # API_ATTACK auth-bypass/IDOR/mass-assignment/BFLA chains) cannot
-        # abort unrelated, still-healthy chat-routed scenarios — only the
-        # remaining not-yet-dispatched direct-HTTP-only scenarios are skipped
-        # (chain_status="target_unreachable"). Latched across passes the same
-        # way self._circuit_open is.
-        self._endpoint_circuit_open = False
         # Auto-discover from SBOM; fall back to provided values
         self._chat_path, self._chat_payload_key, self._chat_payload_list, _discovered_response_key = (
             _discover_chat_config(sbom, chat_path, chat_payload_key, chat_payload_list)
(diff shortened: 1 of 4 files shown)


ganesh-nuguard:
@nikhilpatidar 
This handles temporary server downtime (cold starts, DNS failures, etc.) well. 
But how does it behave when the server is permanently down? Will it still go through all passes?

nikhilpatidar:
Good question. With the per-pass design, a permanently down endpoint still gets probed each pass — up to `_ABORT_THRESHOLD` (3) consecutive `TargetUnavailableError` attempts per pass. After 3 failures in a pass, the remaining direct-HTTP-only scenarios in that pass are skipped, but the next escalation pass starts with a fresh event and re-probes.

So for a 3-pass run with a dead endpoint: 3 probes in pass 1, 3 in pass 2, 3 in pass 3 — bounded at 3×`_ABORT_THRESHOLD` total. This is intentional: a genuinely dead target might recover between passes (container restart, pod rotation, LB failback), and the 3-probe-per-pass cost is acceptable. The prior latched design was worse — one blip permanently killed all `API_ATTACK` scenarios for the entire run.

ganesh-nuguard:
> Good question. With the per-pass design, a permanently down endpoint still gets probed each pass — up to `_ABORT_THRESHOLD` (3) consecutive `TargetUnavailableError` attempts per pass. After 3 failures in a pass, the remaining direct-HTTP-only scenarios in that pass are skipped, but the next escalation pass starts with a fresh event and re-probes.
> 
> So for a 3-pass run with a dead endpoint: 3 probes in pass 1, 3 in pass 2, 3 in pass 3 — bounded at 3×`_ABORT_THRESHOLD` total. This is intentional: a genuinely dead target might recover between passes (container restart, pod rotation, LB failback), and the 3-probe-per-pass cost is acceptable. The prior latched design was worse — one blip permanently killed all `API_ATTACK` scenarios for the entire run.

Okay
can you verify/confirm one more thing
Existing behavior -- when the target is unreachable the status would report as target_unreachable
Now since you are continuing with all pass -- check what status is reported. It might falsely assume all tests ran and report no_findings instead for target_unreachable

nikhilpatidar:
You were right to flag this. I verified the path and found that endpoint-probe skips use `chain_status="target_unreachable"`, but that status was missing from the scan-level health-abort set. As a result, an all-direct-HTTP run could incorrectly fall through to `no_findings`.

Fixed in commit `5ad59f86`: `target_unreachable` is now treated as a health-abort status. Therefore:

- If all scenario records are `target_unreachable` (permanently unreachable direct-HTTP target), the outcome is `aborted_target_unavailable`.
- If direct-HTTP scenarios are `target_unreachable` but chat-routed scenarios complete, the outcome remains `no_findings` because the scan partially ran successfully.
- Findings still take precedence over transport status.

Added regression tests for both all-unreachable and mixed unreachable/completed cases. Validation: 39 targeted tests passed, Ruff clean, and mypy clean.

ganesh-nuguard (approved):
Thanks for taking care

PR #450, opened 2026-08-29
feat(target): guide expired browser session recovery

### LINKED ISSUE #397: Playwright browser-login: automate cookie refresh + CI/binary setup
## Context

PR #396 adds `nuguard target discover-browser`, a Playwright-driven browser-login fallback for target apps behind an interactive redirect/OAuth login (Auth0 Universal Login, SSO) that no static auth config can drive. It captures a session cookie into `target.auth.type: cookie_file` and writes it back into `nuguard.yaml`.

That PR intentionally deferred a few Playwright-specific follow-ups, tracked here:

### 1. Automatic cookie refresh on expiry
`AuthSession` (`nuguard/common/auth.py`) only self-refreshes for `login_flow` auth (`refresh_on_401`) — `cookie_file` has no live-refresh path. Captured session cookies are commonly short-lived (~24h observed against a real target), so a long-running or next-day `behavior`/`redteam` run can start failing mid-run with no automatic recovery today; the user has to notice `auth_failed` and manually re-run `discover-browser`.

Options to explore:
- Detect a 401/403 on `cookie_file` auth mid-run and surface a much clearer "cookie expired — re-run `nuguard target discover-browser --write --yes`" error instead of a generic auth failure.
- Auto-invoke `discover-browser` on expiry if Playwright is installed (would need an opt-in flag, since `behavior`/`redteam` today never write files as a side effect).

### 2. CI / browser-binary setup
`playwright install chromium` is a separate step from `pip install nuguard[browser]` (Playwright ships its own browser-binary downloader). We should:
- Document/automate this in CI workflows that want to exercise `discover-browser` (e.g. a `playwright install --with-deps chromium` step).
- Consider whether the `browser_login` unit tests (`tests/common/test_browser_login_*.py`) need a CI job with the extra installed, vs. relying on `pytest.importorskip` locally (current state — they skip cleanly without the extra).

### 3. Heuristic coverage
The current login/chat-input/send-button heuristics (`nuguard/common/browser_login/heuristics.py`) were tuned against one real Auth0-gated app (kscope/getmosaiccare.com) and caught a couple of subtle bugs in the process (wrong element type for a chat input; an overly-broad `:has-text()` send-button match). As more target apps are run through `discover-browser`, we should keep widening/hardening these heuristics based on real failures rather than guessing upfront — track any app-specific gaps here or as new issues linking back to this one.

Related: #396

fatkobra:
## PR Type

- [ ] Bug fix
- [x] Feature

## What

- Add actionable recovery guidance when `nuguard target verify` receives HTTP 401 or 403 for the default identity using `cookie_file` authentication.
- Direct the user to refresh the saved browser session with:
  `nuguard target discover-browser --write --yes`.
- Preserve the existing authentication diagnostic for other authentication types, identities, and status codes.
- Add focused regression tests for the recovery conditions.
- Update the existing browser-session documentation.
- Add a Chromium-backed browser-login CI job.
- Preserve the existing public API contract gate.
- Require the full test job to wait for lint, public API contract validation, and browser-login coverage.
- Add a workflow regression test preventing any of those required gates from being dropped.

## Why

Browser-captured sessions can expire while their cookie file still exists.

Previously, `target verify` reported that authentication was rejected but did not direct the user to the browser-discovery flow that refreshes the saved session.

The browser-login implementation also depends on the optional Playwright dependency and a separately installed browser binary. The pull-request workflow did not install the `browser` extra or Chromium, allowing browser-dependent behavior to remain unexercised in CI.

While adding that browser coverage, the existing `public-api-contract` dependency was unintentionally removed from the full test job. This revision restores that dependency and adds a regression test so existing PR gates are preserved.

### Expired-session recovery

- Add `_browser_session_refresh_hint` to centralize the recovery conditions and message.
- Print the hint only when:
  - the failed identity is `default`;
  - `target.auth.type` is `cookie_file`; and
  - the response status is HTTP 401 or 403.
- Keep `target verify` non-interactive and side-effect free.
- Retain the existing generic authentication diagnostic.

### Browser CI coverage

- Add an independent `browser_login` job to the PR workflow.
- Install the development, MCP, and browser dependency sets.
- Install Chromium and its required system dependencies through Playwright.
- Launch Chromium against deterministic local HTML and verify that it renders content and executes JavaScript.
- Run browser/discovery-related tests with Playwright and Chromium installed.

### PR gate preservation

The full test job depends on all three required prerequisite jobs:

- `lint`
- `public-api-contract`
- `browser_login`

The browser job remains independent, allowing its result to be reported even when another prerequisite fails.

`tests/test_pr_workflow_dependencies.py` verifies that:

- all three prerequisite jobs exist;
- all three remain listed in `test.needs`; and
- additional future gates can be added without changing the test.

## Scope

This is the single PR for the actionable scope of #397. No second PR is required for the expired-session recovery or browser CI work.

Additional selector heuristics are intentionally excluded unless backed by a documented, reproducible application or fixture. A future concrete selector failure should be tracked separately and include its own regression test.

## Test Steps

Run the workflow dependency and expired-session regression tests:

```bash
uv run pytest -q \
  tests/test_pr_workflow_dependencies.py \
  tests/cli/test_target_auth_diagnostics.py
```

Install and verify the optional browser environment:

```bash
uv sync \
  --dev \
  --extra mcp \
  --extra browser

uv run playwright install \
  --with-deps \
  chromium

NUGUARD_REQUIRE_BROWSER_TESTS=1 \
uv run pytest \
  tests/common/test_browser_runtime.py \
  -q \
  -ra
```

Run browser/discovery-related tests:

```bash
uv run pytest \
  tests/cli \
  tests/common \
  nuguard/common/browser_login \
  --ignore=tests/common/test_browser_runtime.py \
  -k "browser or discover" \
  -q \
  -ra
```

Run package and repository tests:

```bash
uv run pytest nuguard/ -q
make test
```

Run changed-file quality checks and pre-commit:

```bash
uv run ruff format --check \
  nuguard/cli/commands/target.py \
  tests/cli/test_target_auth_diagnostics.py \
  tests/common/test_browser_runtime.py \
  tests/test_pr_workflow_dependencies.py

uv run ruff check \
  nuguard/cli/commands/target.py \
  tests/cli/test_target_auth_diagnostics.py \
  tests/common/test_browser_runtime.py \
  tests/test_pr_workflow_dependencies.py

make precommit-run
git diff --check
```

Run repository-wide lint and type checking:

```bash
make lint
```

## Checks

- [x] Workflow dependency regression test passes.
- [x] The full test job requires `lint`, `public-api-contract`, and `browser_login`.
- [x] Focused expired-session diagnostics pass.
- [x] Chromium runtime smoke test passes without skipping.
- [x] Browser/discovery test selection passes.
- [x] Package-side tests pass.
- [x] `make test` passes.
- [x] `make lint` passes.
- [x] Changed Python files pass Ruff formatting and lint checks.
- [x] `make precommit-run` passes.
- [x] Tests were added for the new behavior, browser runtime, and workflow dependencies.
- [x] User-facing documentation was updated.
- [x] The existing public API contract gate is preserved.

## Other Notes

`target verify` does not automatically open Playwright or modify `nuguard.yaml`. It only provides the command needed to refresh the saved browser session.

The Chromium smoke test uses local HTML and JavaScript only. It does not contact an external application, require credentials, or persist cookies.

The browser CI job is additive. It does not replace lint or public API contract validation.

Repository-wide `make fmt` is intentionally not used for this patch because the current repository baseline reformats hundreds of unrelated files. The Python files changed by this PR are formatted and checked directly.

## Release Note

`nuguard target verify` now suggests rerunning browser discovery when the default saved cookie session may have expired, and browser-login behavior is exercised against Chromium in pull-request CI.

Closes #397

Related: #396


.github/workflows/pr-tests.yml
```diff
@@ -59,13 +61,52 @@ jobs:
...
+            -q \
+            -ra
+
   test:
     name: Run full pytest suite
     runs-on: ubuntu-latest
     timeout-minutes: 30
     needs:
       - lint
-      - public-api-contract
```
ganesh-nuguard, line 68: Are we removing public api contract tests from the PR review?

fatkobra: Good catch. No, removing the public API contract gate was not intended.

The browser CI change replaced the existing `needs` list instead of appending to it. I’ll restore `public-api-contract` and keep `browser_login` as an additional prerequisite, so the full test job depends on all three:

- `lint`
- `public-api-contract`
- `browser_login`

I’ll also update the branch from `develop` and rerun the checks before resolving this thread.

fatkobra: I restored `public-api-contract` as a prerequisite and kept `browser_login` as an additional gate. The full test job now depends on:

- `lint`
- `public-api-contract`
- `browser_login`

I also added a regression test that verifies all three jobs exist and remain in `test.needs`, while still allowing future gates to be added.

ganesh-nuguard (approved):
Thanks for taking care @fatkobra

PR #478, opened 2026-08-31
fix: richer checkov titles + file/line evidence

rangoel-nu:
## Summary
Stacks on #477. Two readability follow-ups for Checkov findings, requested after reviewing that fix's output in a real report:

- Titles showed only the bare check name with no indication of category or affected resource. Now: `[IaC misconfiguration] {check_name} — {resource}`.
- Normal resource checks (e.g. GitHub Actions workflow permissions) never surfaced a file/line pointer into source — only Checkov's logical resource address. Populated the existing, previously-unused `Finding.evidence` field with `file_path:line` and render it as an `Evidence:` line in the markdown report. Omitted when it would duplicate "Affected Components" (the `CKV_SECRET_*` hash-avoidance case from #477).

## Test plan
- [x] `uv run pytest tests/analysis/test_checkov_scanner.py -v` — 8 passed
- [x] `uv run pytest tests/analysis tests/cli -q` — 154 passed
- [x] `uv run ruff check` / `uv run mypy` on both changed files

**Base note:** targets `bug/checkov-fix` (not `develop`) since #477 hasn't merged yet — this diff is additive on top of it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UZ4eUBua9oWeYrWyzsMep8

PR #477, opened 2026-08-31
fix: Checkov title/field parsing and nested asyncio.run() in ATLAS enrichment

rangoel-nu:
## Summary
- Checkov's `_parse_checkov_output` read a nested `"check"` object that doesn't exist in real Checkov JSON, so every finding title collapsed to the bare rule ID and remediation guideline URLs/severity were dropped; `CKV_SECRET_*` findings also leaked a raw secret hash into "Affected Components" instead of a canonical file location.
- ATLAS LLM enrichment silently failed on every `nuguard scan` run because `_run_atlas_native` was called synchronously from within `StaticAnalyzer.analyze()`'s already-running event loop, causing a nested `asyncio.run()` RuntimeError that was swallowed and fell back to static output.
- Bonus: fixed a hardcoded wrong Azure DeepSeek deployment name and a broken YAML indentation (`eval_llm` silently parsed as `None`) in the juice-shop Azure test config, found while diagnosing a related Azure auth failure.

## Test plan
- [x] `uv run pytest tests/analysis -q` — 26 passed
- [x] `uv run pytest tests/analysis/test_checkov_scanner.py -v` — 6 passed (new)
- [x] `uv run ruff check nuguard/analysis/plugins/checkov_scanner.py nuguard/analysis/static_analyzer.py tests/analysis/test_checkov_scanner.py`
- [x] `uv run mypy nuguard/analysis/plugins/checkov_scanner.py nuguard/analysis/static_analyzer.py`

Closes #475
Closes #476

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UZ4eUBua9oWeYrWyzsMep8

PR #479, opened 2026-08-31
fix: dedup checkov findings for the same rule + affected component

rangoel-nu:
## Summary
Stacks on #478 (stacks on #477). Fixes visually duplicated Checkov finding blocks in reports, e.g. `CKV_AWS_150`/`CKV_AWS_91` each appearing twice under `aws_lb.juice_shop` in `tests/apps/owasp-juice-shop/reports/juice-shop-analysis.md`.

**Checked for a checkov config option first** — none exists (`checkov --help` has no dedup/uniqueness flag; `--compact` only suppresses code blocks in CLI text output). `--skip-path`/`--download-external-modules` solve a different problem (stale `.terraform/` module caches) and don't apply here.

**Root cause**: this SBOM's IaC-typed nodes carry no `source_path`, so checkov runs exactly once against the whole repo — not once per overlapping path. The "duplicates" are checkov legitimately reporting the same Terraform resource *address* defined in two different files in the repo (`terraform/networking.tf` and `infrastructure/terraform/networking.tf` both define `aws_lb.juice_shop`). Since the report groups by bare resource-address text with no file qualifier, these two distinct findings render as repeated, identical-looking blocks.

**Fix**: added `_dedupe_findings()`, applied once across all collected findings in `CheckovScannerPlugin.run()`. Collapses on `(rule_id, affected component)` — rather than dropping the second occurrence outright, merges the `evidence` (file:line) values from every duplicate so a reader still sees every affected location instead of losing information.

## Test plan
- [x] `uv run pytest tests/analysis/test_checkov_scanner.py -v` — 11 passed (3 new dedup tests)
- [x] `uv run pytest tests/analysis tests/cli -q` — 157 passed
- [x] `uv run ruff check` / `uv run mypy` on the changed file

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UZ4eUBua9oWeYrWyzsMep8

### DIFF
--- a/nuguard/analysis/plugins/checkov_scanner.py
+++ b/nuguard/analysis/plugins/checkov_scanner.py
@@ -90,6 +90,7 @@ class CheckovScannerPlugin(AnalysisPlugin):
         all_findings: list[dict[str, Any]] = []
         for path in iac_paths:
             all_findings.extend(_run_checkov(binary, path, config))
+        all_findings = _dedupe_findings(all_findings)
 
         status = "warning" if all_findings else "ok"
         message = (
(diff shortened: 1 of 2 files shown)


ganesh-nuguard:
@rangoel-nu you have created PR from one feature branch to other feature branch - fyi

rangoel-nu:
Superseded by #497, which combines and rebases this stack onto latest develop.

PR #484, opened 2026-09-02
feat(websocket): detect WebSocket chat endpoints (Layer 1)

### LINKED ISSUE #486: Support for websockets
# WebSocket Support — Implementation Plan

**Related issue:** NuGuardAI/NuGuard-app#1117 (chat endpoint detection gaps)
**Related PR:** NuGuardAI/nuguard#385 (gap #5 — detection only, deferred)

WebSocket support requires changes across four layers. Each layer depends on the one above it being complete first.

---

## Background

Some agentic apps expose their chat interface exclusively over WebSocket:
- **LangGraph Cloud** — streaming agent responses over `wss://`
- **OpenAI Realtime API** — bidirectional audio/text over WebSocket
- **Custom agent SDKs** — multi-turn conversation managed by the server over a persistent socket

WebSocket differs from HTTP in one critical way: the connection is **persistent and stateful**. A single `wss://` connection stays open for the entire conversation, and the server can push messages at any time. This affects all four layers below.

---

## Layer 1 — Detection (`nuguard/common/endpoint_probe.py`)

**Current state:** detection only does HTTP POST. A GET to a WS endpoint returns `101 Switching Protocols` or `426 Upgrade Required`, both of which cause the probe to skip the path silently.

**What to add:**

1. **SBOM detection (zero-I/O):** In `_sbom_post_paths()` and `discover_chat_candidates_from_sbom()`, check if any API_ENDPOINT node's URL starts with `ws://` or `wss://`, or if the path contains `/ws`, `/websocket`, `/socket.io`, `/realtime`. Return these with a `__websocket__` payload key marker (same pattern as `__adk__` for Google ADK).

2. **Live detection probe:** In `_detect_chat_endpoint()`, add an Option N that sends an HTTP GET with `Connection: Upgrade` and `Upgrade: websocket` headers to WS-candidate paths. A `101` or `426` response confirms the path is a WS endpoint. Return `(path, "__websocket__", False)`.

3. **Fallback path list:** Add `/ws`, `/ws/chat`, `/socket`, `/realtime` to the standard fallback paths in `probe_chat_endpoints()`.

**Files:** `nuguard/common/endpoint_probe.py`

---

## Layer 2 — Target Client (`nuguard/common/target_client_builder.py`, `nuguard/redteam/target/client.py`)

**Current state:** `TargetAppClient` only knows HTTP. Its `send(message, session)` method does `await client.post(path, content=json.dumps(body))`.

**What to add:**

1. **New dependency:** Add `websockets>=12.0` to `pyproject.toml` (optional dependency or required). The `websockets` library provides `websockets.connect()` for async WS client connections.

2. **`WebSocketTargetClient` class:** A new class (or mode in `TargetAppClient`) that:
   - Opens a `websockets.WebSocketClientProtocol` in `__aenter__` (the existing context-manager entry)
   - Sends auth via the initial Upgrade request headers (`Authorization: Bearer {token}`) OR as a first JSON message after connecting (target-dependent)
   - Implements `send(message, session)` by sending a JSON frame and awaiting the response frame
   - Closes the socket in `__aexit__`

3. **`build_target_app_client()` factory:** When `payload_key == "__websocket__"`, instantiate `WebSocketTargetClient` instead of the HTTP client. Pass the `base_url` with scheme converted from `http(s)://` to `ws(s)://`.

4. **Message format:** WS messages are typically `{"role": "user", "content": "..."}` (OpenAI Realtime style) or a flat `{"message": "..."}` — this should be configurable via `chat_payload_key` and `chat_payload_list` exactly as HTTP is.

**Files:**
- `nuguard/common/target_client_builder.py`
- `nuguard/redteam/target/client.py` (or a new `nuguard/common/ws_client.py`)
- `pyproject.toml`

---

## Layer 3 — Session Bootstrap (`nuguard/common/session_resolver.py`, `nuguard/common/bootstrap.py`)

**Current state:** `bootstrap_auth_runtime()` sends a test POST to the chat endpoint to confirm the target is reachable and to trigger auth. This doesn't work for WS.

**What to add:**

1. **WS bootstrap path:** When `chat_path` resolves to a WS endpoint, `bootstrap_auth_runtime()` should attempt a WebSocket connect (with auth headers) instead of a POST. A successful `101` handshake with no immediate close counts as "target is reachable and auth is accepted."

2. **Auth over WS — two patterns to support:**
   - **Header auth (most common):** Token passed as `Authorization: Bearer {token}` in the Upgrade request headers. This works with the existing `AuthSession.headers()` mechanism if the WS client forwards them on connect.
   - **First-message auth (some platforms):** After connecting, the server expects a first message like `{"type": "auth", "token": "..."}` before accepting chat messages. This requires a `ws_auth_message` config option.

3. **Cold-start retry:** The existing 3-retry / 2-4-8s backoff for cold containers should apply to WS connects the same as HTTP probes.

**Files:**
- `nuguard/common/session_resolver.py`
- `nuguard/common/bootstrap.py`

---

## Layer 4 — Runners (`nuguard/behavior/runner.py`, `nuguard/redteam/executor/orchestrator.py`)

**Current state:** Both runners create a `TargetAppClient` per scenario and call `send()` in a loop. They assume HTTP semantics (stateless, reconnect is free).

**What to add:**

1. **Session lifetime:** For WS, the socket MUST stay open across all turns of a scenario. The current `async with build_target_app_client(...) as client:` per-scenario pattern already handles this correctly IF `WebSocketTargetClient.__aenter__` opens the socket and `__aexit__` closes it. No runner change needed for this if the client abstraction is clean.

2. **Server-push messages:** Some WS agents push unsolicited messages (typing indicators, partial tokens) between turns. The `send()` implementation needs to drain/ignore these until a "complete" message arrives. A common pattern is waiting for a message with `"type": "response.complete"` or similar. This may require a `ws_response_complete_key` config option.

3. **Connection drops mid-scenario:** WS connections can drop unexpectedly. The runner should catch `websockets.ConnectionClosed` and either reconnect (re-opening at the same scenario state, if the server supports it) or mark the scenario as failed. Currently HTTP requests just fail and retry; WS needs explicit reconnect logic.

4. **No change needed for:** scenario selection, finding generation, scoring, reporting — these are all above the `send()` abstraction layer.

**Files:**
- `nuguard/behavior/runner.py`
- `nuguard/redteam/executor/orchestrator.py`

---

## NuGuard-app wiring

Once the `nuguard` library supports WebSocket via the `__websocket__` marker:

1. **`executor_utils.py`:** `run_prescan_discovery` already calls `probe_chat_endpoints` — no change needed if the marker is transparent to the platform layer.
2. **`schemas.py`:** No new fields needed if WebSocket is auto-detected. Optionally add `ws_auth_message` to `RunBehaviorRequest` / `RunRedTeamRequest` for the first-message auth pattern.
3. **DB migration:** Possibly add `ws_auth_message` column to the runs table.

---

## Implementation order

1. Layer 1 — Detection (can be done without `websockets` dependency; just HTTP 426 detection)
2. Layer 2 — `WebSocketTargetClient` (add `websockets` dependency; most of the work)
3. Layer 3 — Bootstrap (needs Layer 2 complete)
4. Layer 4 — Runner edge cases (connection drops, server-push draining)
5. NuGuard-app wiring

---

## Open questions before starting

- Which specific WS agents are in scope for the first release? (LangGraph Cloud / OpenAI Realtime / both / other?)
- Is header auth sufficient or do we need to support first-message auth for any known target?
- Should `websockets` be a hard dependency or optional (only pulled in when a WS endpoint is detected)?
- What is the expected message format for WS targets — flat JSON, OpenAI Realtime frames, or both?

ganesh-nuguard:
Summary:
Wires up the SBOM adapter's WEBSOCKET-method endpoint nodes (added in commit dd8a4f5d, `fastapi_adapter.py`) into the chat-endpoint discovery pipeline described in `imp_docs/websocket_support.md` (Layer 1 — Detection). This PR only covers Layer 1; Layers 2-4 (target client, session bootstrap, runner support) will land as follow-up commits on this same branch/PR.

- `_sbom_websocket_paths()`: zero-I/O extraction of WS paths from SBOM API_ENDPOINT nodes with `method == "WEBSOCKET"`
- `discover_chat_candidates_from_sbom()`: now yields `__websocket__`-keyed candidates for WS nodes (skips path-param routes)
- `probe_chat_endpoints()` / `_detect_chat_endpoint()`: new live-probe option that sends an HTTP GET with `Connection: Upgrade`/`Upgrade: websocket` headers against SBOM-declared and fallback WS paths (`/ws`, `/ws/chat`, `/socket`, `/realtime`); a 101 or 426 response confirms a WebSocket endpoint and returns `ProbeResult(path, "__websocket__", False)`

Validation:
- Lint: `ruff check nuguard/common/endpoint_probe.py nuguard/common/tests/test_endpoint_probe_websocket.py tests/common/test_endpoint_probe.py` → All checks passed!
- Types: `mypy nuguard/common/endpoint_probe.py` → Success: no issues found in 1 source file
- Tests: `python -m pytest nuguard/common/tests/ tests/common/ nuguard/redteam/tests/test_endpoint_probe_circuit_breaker.py -q` → 403 passed, 2 skipped

Closes #486


ganesh-nuguard:
Commit: 9f9345df

Summary:
Added Layer 2 of the WebSocket support plan (imp_docs/websocket_support.md) — WebSocketTargetClient in nuguard/redteam/target/ws_client.py, with build_target_app_client() branching to it when the discovered payload key is the __websocket__ marker from Layer 1 detection. Supports header auth, first-message auth (ws_auth_message), server-push draining via ws_response_complete_key, transparent reconnect after a dropped connection, and the existing TargetUnavailableError circuit breaker. Added websockets>=12.0 to pyproject.toml.

Validation:
Lint: ruff check nuguard/redteam/target/ws_client.py nuguard/redteam/tests/test_ws_target_client.py nuguard/common/target_client_builder.py tests/common/test_target_client_builder.py pyproject.toml -> All checks passed!
Types: mypy nuguard/redteam/target/ws_client.py nuguard/common/target_client_builder.py -> Success: no issues found in 2 source files
Tests: python -m pytest nuguard/redteam/tests/ tests/common/ nuguard/common/tests/ -q -> 797 passed, 2 skipped, 2 pre-existing failures unrelated to this change (test_llm_client_stream_retry.py order-dependent pollution, reproduces identically on develop without these changes)

ganesh-nuguard:
Commit: 4976fb20

Summary:
Added Layer 3 of the WebSocket support plan (imp_docs/websocket_support.md) — WebSocket-aware session bootstrap. AuthBootstrapper now opens a WS handshake (instead of an HTTP POST) for the pre-run connectivity/auth health check when the target is a WebSocket endpoint: is_websocket/ws_auth_message options, a new _probe_websocket_once() mapping 401/403 handshake rejections to auth_failed and everything else to target_unavailable (reusing the existing cold-start retry/backoff), and support for both header auth (Upgrade request headers) and first-message auth (ws_auth_message). resolve_target_session() now detects a WS target zero-I/O from the SBOM before calling bootstrap, so bootstrap picks the right probe strategy up front. Renamed ws_client._to_ws_url -> to_ws_url since it's now shared with bootstrap.py.

Validation:
Lint: ruff check nuguard/common/bootstrap.py nuguard/common/auth_runtime.py nuguard/common/session_resolver.py nuguard/redteam/target/ws_client.py tests/common/test_bootstrap_websocket.py tests/common/test_session_resolver.py -> All checks passed!
Types: mypy nuguard/common/bootstrap.py nuguard/common/auth_runtime.py nuguard/common/session_resolver.py nuguard/redteam/target/ws_client.py -> Success: no issues found in 4 source files
Tests: python -m pytest tests/common/ nuguard/common/tests/ nuguard/redteam/tests/ -q -> 803 passed, 2 skipped, 2 pre-existing failures unrelated to this change (same test_llm_client_stream_retry.py order-dependent pollution noted on the previous commit, reproduces identically on develop without these changes)

ganesh-nuguard:
Commit: 7f3cc8d7

Summary:
Deep-dive review of the actual execution paths (per user request) found two real gaps the earlier layers missed:

1. Neither BehaviorRunner nor RedteamOrchestrator call resolve_target_session() — they each duplicate their own bootstrap logic inline, so Layer 3's WS-aware AuthBootstrapper was never actually reachable from a real run. Fixed by factoring the WS zero-I/O pre-detection into a new sbom_indicates_websocket() helper (endpoint_probe.py) and calling it from both runner.py._build_client() and orchestrator.py before bootstrap_auth_runtime().
2. Found and fixed a probe_chat_endpoints() bug where hint_path (Option B — explicit target_endpoint configured) cleared ws_paths, silently disabling WS detection for a user-configured WS endpoint.
3. Found a WebSocketTargetClient/TargetAppClient interface-parity gap: runner.py/executor.py call set_chat_endpoint(), set_path_param(), update_default_headers(), invoke_endpoint(), and .chat_path unconditionally on the shared client regardless of type. Missing these on WebSocketTargetClient would crash direct-HTTP attack scenarios (IDOR/BFLA/mass-assignment) the first time they ran against a WS-chat target. Added all of them, including invoke_endpoint() over a separate lazily-created httpx client (WS-chat targets often still expose plain REST routes for non-chat resources).
4. Added a TargetClient = TargetAppClient | WebSocketTargetClient type alias and widened client parameter types across discovery.py, endpoint_preflight.py, executor.py, guided_executor.py, turn_helpers.py, and tree_explorer.py to match build_target_app_client()'s real return type (mypy-only fixes, no behavior change).

Validation:
Lint: ruff check <all changed files> -> All checks passed!
Types: mypy nuguard/ (whole package) -> Success: no issues found in 618 source files
Tests: python -m pytest nuguard/redteam/tests/ nuguard/behavior/tests/ nuguard/common/tests/ tests/common/ tests/redteam/ -q -> 2178 passed, 7 skipped

ganesh-nuguard:
Commit: ad730f8a

Summary:
Live-tested the whole WS support pipeline against a real, running app for the first time: cloned bytedance/deer-flow, generated its SBOM (495 nodes, confirmed one real WEBSOCKET-method node: /api/threads/{thread_id}/browser/stream), installed its backend deps and ran it locally (uv sync + uvicorn app.gateway.app:app), then ran 'nuguard target verify' against it end-to-end using the tests/local/nuguard-deer-flow.yaml config.

That surfaced a real bug: discover_chat_candidates_from_sbom() was unconditionally dropping any WEBSOCKET-method node whose path had a {param} placeholder, instead of penalizing it like the equivalent HTTP branch does. deer-flow's actual WS route is exactly this shape, so it was invisible to discovery entirely. Fixed to penalize (-5, matching the HTTP path-param penalty) instead of dropping, so it still surfaces as a fallback candidate and can be selected via path_param_sources resolution when nothing parameter-free exists.

For deer-flow specifically, the live run confirms the *correct* outcome: two real HTTP chat endpoints outrank the WS browser-stream route (which is a secondary live-preview feature, not deer-flow's actual chat interface) — 'nuguard target verify' picked the top HTTP candidate and got as far as a CSRF-protected 403, i.e. the SBOM-driven bootstrap pipeline (Layers 1 and 3) ran correctly against a real app end-to-end.

Validation:
Lint: ruff check nuguard/common/endpoint_probe.py tests/common/test_endpoint_probe.py -> All checks passed!
Types: mypy nuguard/common/endpoint_probe.py -> Success: no issues found in 1 source file
Tests: python -m pytest tests/common/ nuguard/common/tests/ -q -> 406 passed, 2 skipped, 2 pre-existing failures unrelated to this change (same test_llm_client_stream_retry.py order-dependent pollution noted on earlier commits)
Live: nuguard sbom generate + nuguard target verify against a locally running deer-flow instance (tests/local/nuguard-deer-flow.yaml, gitignored per repo convention)

ganesh-nuguard:
Commit: 48d01376

Summary:
Switched test targets: deer-flow's chat interface turned out to be HTTP (its WS route is a secondary browser-preview feature), so picked NuGuardAI/voicelive-api-salescoach-demo instead — a Flask + flask_sock app whose /ws/voice route IS the real chat/voice interface. This immediately surfaced a bigger gap than the deer-flow one: FlaskAdapter had zero support for flask_sock's @sock.route(...) — the flask_sock equivalent of FastAPI's @app.websocket(...) (fixed earlier in dd8a4f5d). The route was completely invisible to the SBOM, not just mis-scored.

Fix in nuguard/sbom/adapters/python/flask_adapter.py:
- Track Sock(app) instantiations (mirrors existing Flask()/Blueprint() tracking) so .route(...) calls on that variable are tagged method=WEBSOCKET.
- Resolve module-level string constants used as decorator path args (WEBSOCKET_ENDPOINT = "/ws/voice" then @sock.route(WEBSOCKET_ENDPOINT)) since plain AST has no constant folding — this is exactly how flask_sock apps declare their route path.
- Skip HTTP-only chat_payload_key body inference for WS handlers (they use ws.receive(), not request.json).

Live end-to-end validation (config/report under tests/local/nuguard-voicelive-salescoach.yaml, gitignored per repo convention, repo cloned as a sibling):
1. Installed backend deps, ran the real Flask app locally (no Azure creds needed — it degrades gracefully)
2. SBOM regenerated: WEBSOCKET /ws/voice now detected (API_ENDPOINT count 4→9, other paths improved too from the constant-resolution side effect)
3. discover_chat_candidates_from_sbom() selects /ws/voice as the sole candidate
4. 'nuguard target verify' completes a real WS handshake against the live app — status=ok
5. Direct WebSocketTargetClient.send() call over the real socket: sent a JSON frame, received a real response frame back from the live server (its actual 'no API key configured' error path) — full connect→send→recv→close round trip confirmed against an unmodified third-party app

Validation:
Lint: ruff check nuguard/sbom/adapters/python/flask_adapter.py nuguard/sbom/tests/test_flask_adapter.py -> All checks passed!
Types: mypy nuguard/sbom/adapters/python/flask_adapter.py -> Success: no issues found in 1 source file
Tests: python -m pytest nuguard/sbom/tests/ -q -> 1389 passed, 1 skipped

ganesh-nuguard:
Commit: 8f3e6267

Summary:
Found and fixed the e2e (gemini-auto) CI failure — a real regression from our WebSocket detection changes, not a flaky test.

Root cause: _probe_websocket_upgrade() treated ANY HTTP 426 (Upgrade Required) response as confirmation of a WebSocket endpoint. gemini-auto's /api/agent/chat is a plain HTTP POST JSON endpoint deployed on Cloud Run — a GET probe to it returned a bare 426 (a generic wrong-method/protocol error, unrelated to WebSocket support), which our probe misread as __websocket__. The subsequent real WS handshake attempt then correctly failed ('missing Sec-WebSocket-Accept header'), retried 3x with backoff, and the whole redteam run aborted with 'target is unreachable', which failed the prepublish gate check.

Fix: only accept 101 (Switching Protocols) as unconditional confirmation. A 426 is now only accepted when its Upgrade response header explicitly names 'websocket' (RFC 9110 section 15.5.22) — the correct way a server signals 'I don't support this protocol, but I would upgrade to X'. A bare 426 with no such header is no longer treated as a WS signal.

Validation:
Lint: ruff check nuguard/common/endpoint_probe.py nuguard/common/tests/test_endpoint_probe_websocket.py -> All checks passed!
Types: mypy nuguard/common/endpoint_probe.py -> Success: no issues found in 1 source file
Tests: python -m pytest tests/common/ nuguard/common/tests/ nuguard/redteam/tests/ -q -> 809 passed, 2 skipped, 2 pre-existing unrelated failures (same order-dependent test_llm_client_stream_retry.py pollution noted on earlier commits). Added test_probe_ignores_bare_426_with_no_upgrade_header() regression test that reproduces the exact gemini-auto shape (POST-only endpoint answering GET with a bare 426).

ganesh-nuguard:
Commit: bf7f9d21

Summary:
The previous fix (8f3e6267) only tightened the 426 branch — the e2e (gemini-auto) test was still failing because a bare 101 status was still trusted unconditionally. Root cause: gemini-auto's Cloud Run deployment answers the live upgrade probe on /api/agent/chat (a plain HTTP POST-only endpoint) with a genuine 101 status. Some serverless/reverse-proxy front-ends complete a generic HTTP/1.1 Upgrade handshake for any path regardless of whether the application itself implements WebSocket there.

Fix: validate Sec-WebSocket-Accept per RFC 6455 4.2.2 — base64(sha1(key + '258EAFA455E4B4CE-C5B58384CD9835B4')). Only a 101 response with that exact computed value now counts as WS confirmation. This is exactly what the real 'websockets' library validates during a genuine handshake, which is also why the bootstrap step was already failing with 'missing Sec-WebSocket-Accept header' — we just weren't checking for that signal in the cheaper live probe.

Extracted compute_websocket_accept() so probe code and tests share the same RFC computation. Updated the existing tests to mock a spec-compliant response via a request-aware side_effect, and added a new regression test reproducing the exact gemini-auto shape (bare 101, no Accept header).

Validation:
Lint: ruff check nuguard/common/endpoint_probe.py nuguard/common/tests/test_endpoint_probe_websocket.py -> All checks passed!
Types: mypy nuguard/common/endpoint_probe.py -> Success: no issues found in 1 source file
Tests: python -m pytest tests/common/ nuguard/common/tests/ nuguard/redteam/tests/ -q -> 810 passed, 2 skipped, 2 pre-existing unrelated failures (same order-dependent test_llm_client_stream_retry.py pollution noted on earlier commits)

ganesh-nuguard:
@copilot resolve the merge conflicts in this pull request

ganesh-nuguard:
Commit: c1c03fa5

Summary:
Fixed the lint/type-check workflow failure. Mypy reported that the shared warmup callers passed retry_transient=True to WebSocketTargetClient.send(), but the WS client did not expose that interface option. Added the retry_transient keyword and bounded reconnect/backoff retries for transient WS failures during pre-run warmup. Normal scenario sends remain unchanged.

Validation:
- mypy nuguard/redteam/target/ws_client.py nuguard/redteam/executor/orchestrator.py nuguard/common/session_resolver.py -> Success: no issues found
- ruff check nuguard/redteam/target/ws_client.py -> All checks passed
- python -m pytest nuguard/redteam/tests/test_ws_target_client.py -q -> 10 passed

